PR #1, opened 2017-08-25
finished!!!

WizrdJim:
done just had to force it into submission

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -1,6 +1,8 @@
 const bodyParser = require('body-parser');
 const express = require('express');
 const session = require('express-session');
+const User = require('./user');
+const bcrypt = require('bcrypt');
 
 const STATUS_USER_ERROR = 422;
 const BCRYPT_COST = 11;
@@ -14,6 +16,7 @@ server.use(session({
 
 /* Sends the given err, a string or an object, to the client. Sets the status
  * code appropriately. */
+
 const sendUserError = (err, res) => {
   res.status(STATUS_USER_ERROR);
   if (err && err.message) {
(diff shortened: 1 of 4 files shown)


src/server.js
```diff
@@ -23,12 +30,103 @@ const sendUserError = (err, res) => {
...
+    req.user = user;
+  });
+});
+
 // TODO: implement routes
+server.post("/users", (req, res) => {
+  // getting the username and password from a user
+  const { username, password } = req.body;
+  // Check if the password is empty
+  if (password === "") {
```
seanchen1991: Here you can just do `if (!password)`, which will check for empty strings among other non-truthy values.

src/server.js
```diff
@@ -23,12 +30,103 @@ const sendUserError = (err, res) => {
...
+        return;
+      }
+      res.json(user);
+    });
+  });
+});
+server.post("/login", (req, res) => {
+  const { username, password } = req.body;
+  const sess = req.session;
+  if (!password || password === "") {
```
seanchen1991: Same deal here. `if (!password)` is already checking for `password === ""`

src/server.js
```diff
@@ -23,12 +30,103 @@ const sendUserError = (err, res) => {
...
+      return;
+    }
+    if (!user) {
+      sendUserError("Please input valid info", res);
+      return;
+    }
+    if (bcrypt.compareSync(password, user.passwordHash)) {
+      sess.LoginId = user.id;
+      res.json({ success: true });
+      return;
```
seanchen1991: Whenever you call `res`, i.e. with `res.json` or `res.send`, there's an implicit `return`, so you don't need to call `return` here since you're already calling `res.json`

PR #2, opened 2017-08-28
Auth sprint completed

loyslane:
All codes are within our individual Lois/patrick folders.


Lois/src/server.js
```diff
@@ -0,0 +1,116 @@
...
+    return;
+  }
+  if (!password) {
+    sendUserError('Please enter a password!', res);
+    return;
+  }
+  User.findOne({ username })
+    .exec()
+    .then((user) => {
+      if (user.length < 1) {
```
seanchen1991: Here you can do `if (!user)` since you want to check that the user array is empty, and an empty array is falsey.

Lois/src/server.js
```diff
@@ -0,0 +1,116 @@
...
+  if (err && err.message) {
+    res.json({ message: err.message, stack: err.stack });
+  } else {
+    res.json({ error: err });
+  }
+};
+
+// TODO: implement routes
+const userAuthMiddleware = (req, res, next) => {
+  if (req.session.user === undefined) {
```
seanchen1991: Here you can just check `if (!req.session.user)` instead of explicitly checking if it is `undefined`.

Lois/src/server.js
```diff
@@ -0,0 +1,116 @@
...
+    sendUserError('Must be logged in!', res);
+  } else {
+    req.user = req.session.user;
+    next();
+  }
+};
+
+server.use('/restricted', (req, res, next) => {
+  if (req.session.user === undefined) {
```
seanchen1991: Same here.

PR #3, opened 2017-08-28
most routes functional

bdurb:


src/server.js
```diff
@@ -23,8 +25,55 @@ const sendUserError = (err, res) => {
   }
 };
 
+const validateNameAndPass = (req, res, next) => {
+  const username = req.body.username;
+  const password = req.body.password;
+  if (!username || !password) {
+    res.status(STATUS_USER_ERROR);
+    res.json({ error: 'Must provide username and password' });
+    return;
```
seanchen1991: Calling `res.json` and other `res` functions like `res.send` performs an implicit `return`, so calling `return` afterwards is extraneous.

src/server.js
```diff
@@ -23,8 +25,55 @@ const sendUserError = (err, res) => {
...
+  next();
+};
+
 // TODO: implement routes
 
+// User signup
+server.post('/users', validateNameAndPass, (req, res) => {
+  bcrypt.hash(req.password, BCRYPT_COST, (err, hash) => {
+    if (err) {
+      sendUserError();
```
seanchen1991: Make sure you're actually passing the `err` into the `sendUserError()` method, otherwise it won't print a helpful error.

src/server.js
```diff
@@ -23,8 +25,55 @@ const sendUserError = (err, res) => {
...
+  });
+});
+
+// User login
+server.post('/log-in', validateNameAndPass, (req, res) => {
+  // here we hash the password that the client inputs
+  bcrypt.hash(req.password, BCRYPT_COST, (err, hash) => {
+    req.username = req.session.username;
+    if (err) {
+      sendUserError();
```
seanchen1991: Same thing here. Make sure to pass `err` to `sendUserError()`.

src/server.js
```diff
@@ -23,8 +25,55 @@ const sendUserError = (err, res) => {
...
+    } else {
+      // here we search for a user with the same username that the client inputs
+      User.findOne({ username: req.username }, (err, foundUser) => {
+      if (foundUser.passwordHash === hash) {
+        res.json({ success: true });
+      } else {
+        sendUserError();
```
seanchen1991: Same thing here.

PR #4, opened 2017-08-29
Auth working

wlharvey4:


src/user.js
```diff
@@ -10,6 +10,16 @@ mongoose.connect('mongodb://localhost/users', { useMongoClient: true });
...
   // TODO: fill in this schema
+  username: {
+    type: String,
+    required: true,
+    unique: true,
+  },
+  passwordHash: {
+    type: String,
+    required: true,
+    unique: true,
```
seanchen1991: I wouldn't actually enforce passwords to be unique. It's very possible for two people to pick the same password, especially since most people pick from a pool of very common passwords.

PR #5, opened 2017-08-30
Finished Auth(forgot to pull request)

jeshsy:


src/server.js
```diff
@@ -23,12 +26,86 @@ const sendUserError = (err, res) => {
...
+      return;
+    }
+    if(!user) {
+      sendUserError('username required', res);
+      return;
+    }
+    if(bcrypt.compareSync(password, user.passwordHash)) {
+      session.login = user.username;
+      res.json({ success: true });
+      return;
```
seanchen1991: No need to explicitly call `return` here after calling `res.json`, or after any `res` method like `res.send`.

src/server.js
```diff
@@ -23,12 +26,86 @@ const sendUserError = (err, res) => {
...
 
 // TODO: add local middleware to this route to ensure the user is logged in
-server.get('/me', (req, res) => {
+const localMiddleware = (req, res, next) => {
+  const session = req.session;
+  if (!session.login) {
+    sendUserError('Login required', res);
+    return;
+  }
+  User.find({ username: session.login }, (err, user) => {
```
seanchen1991: Here, I would use `User.findOne` or `User.findById` instead of just `find`. This works fine, but I think it's good to use the slightly more explicit method.

src/server.js
```diff
@@ -23,12 +26,86 @@ const sendUserError = (err, res) => {
...
+    if (err) {
+      sendUserError(err, res);
+      return;
+    }
+    req.user = user[0];
```
seanchen1991: Then here, you would just do `req.user = user;` if we use `User.findOne` or `User.findById`.

PR #116, opened 2018-04-17
Jonathan Reeves WIP Auth

RedHoodJT1988:


src/server.js
```diff
@@ -23,10 +33,78 @@ const sendUserError = (err, res) => {
   }
 };
 
+
+const testUsername = function (req, res, next) {
+  const { username } = req.body;
+
+  if (!username || username.trim() === '') {
```
randallwhitlock: A typical convention used when referring to things with falsy values, such as empty strings, is to check for falsiness or truthiness.  So you could just use `if (!username || !username.trim())` instead of using the === evaluation.

PR #123, opened 2018-04-17
abraham andujo

foxtara1:


src/server.js
```diff
@@ -1,34 +1,75 @@
...
-const sendUserError = (err, res) => {
-  res.status(STATUS_USER_ERROR);
-  if (err && err.message) {
-    res.json({ message: err.message, stack: err.stack });
-  } else {
-    res.json({ error: err });
-  }
-};
+server.post('/users', (req, res) => {
+    const { username, password } = req.body;
```
lefrenk: Awesome! Love that you're using object destructuring.

src/server.js
```diff
@@ -1,34 +1,75 @@
...
 
-// TODO: implement routes
+    if (username && password) {
+        bcrypt.hash(password, BCRYPT_COST, (err, hash) => {
+            if (err) {
+                res.status(STATUS_USER_ERROR).json(err);
```
lefrenk: Make _sure_ you `return` in this `if` statement so that the rest of the function stops its execution.

src/server.js
```diff
@@ -1,34 +1,75 @@
...
+            }
+            const user = new User({ username, passHash: hash });
+            user.save()
+                .then((item) => {
+                    res.status(200).json(item);
```
lefrenk: Status code should be `201`

src/server.js
```diff
@@ -1,34 +1,75 @@
...
+                })
+                .catch((error) => {
+                    res.status(STATUS_USER_ERROR).json(err);
```
lefrenk: `res.json(err)` doesn't actually send the entire Error Object to the client. You can do something like `res.json({ message: err.message })`

The Error Object has a `message` property on it. Read more about the Error Object: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

lefrenk:
Hey Abe. Looks like you've got a good grasp of server-side authentication ✔️

The only issue I had was running the `npm test`s. This assignment was expecting you to use the TDD (test driven development) process of developing software.

Also, whenever possible, use ES6 syntax that we've learned in class.

Minor suggestion: Look into using async/await instead of regular Promises.

PR #132, opened 2018-05-14
Kaitlyn Flynn - Auth Project (WIP)

kaitlynflynn:
**Download Project Files and Install Dependencies**

- [x] Fork and Clone this repository.
- [x] CD into the folder where you cloned the repository.
_Do your magic!_

**Implement Requirements**

- [x] Use mongoose middleware to implement password hashing.
- [x] Use local express middleware and mongoose methods to implement password verification.
- [x] Use Postman to test the API as you work through the exercises.

**Design and build a set of endpoints listed below:**
- [x] POST /api/register
- [x] POST /api/login
- [x] GET /api/users

Method | Endpoint | Description
-- | -- | --
POST | /api/register | Creates a user using the information sent inside the body of the request. Hash the password before saving the user to the database.
POST | /api/login | Use the credentials sent inside the body to login the user. On successful login, create a new session for the user and send back a 'Logged in' message and a cookie that contains the user id. If login fails, repond with the correct status code and the message: 'You shall not pass!'
GET | /api/users | If the user is logged in, respond with an array of all the users contained in the database. If the user is not logged in repond with the correct status code and the message: 'You shall not pass!'. Use this endpoint to verify that the password is hashed before it is saved.

**Stretch Problem**
Write a piece of global middleware that ensures a user is logged in when accessing any route prefixed by /api/restricted/. For instance, /api/restricted/something, /api/restricted/other, and /api/restricted/a should all be protected by the middleware; only logged in users should be able to access these routes.
Build a React application that implements components to register, login and view a list of users.


jtla3:
* Great commits and detailed
* Code is very clean and nice to read
* Able to hash password with bcyrpt 
* Able to understand sessions and cookies
* Able to understand web hooks
* Routes work in postman
* MVP is done
* Finally!! got gitignore to work

PR #155, opened 2018-05-14
CS9 Brandon Alexander Auth

BigBricks:


jtla3:
* Great commits but use more detailed naming conventions
* Code is very clean and very readable
* Able to hash password with bcyrpt 
* Able to understand sessions and cookies
* Able to understand web hooks
* Routes work in postman
* MVP is done and stretch goal of custom middleware and front-end (React)

PR #174, opened 2018-05-16
David and Yasin 

yshuman1:


alexfigliolia:
Great work Yasin. You and David have conquered node with flying colors. When connecting to your server via ajax, please don't forget to use Cors so that your requests don't get blocked! Excellent work and please extend my praise to David!

PR #177, opened 2018-06-11
Danny Vail

d-vail:
# Authentication Project

### Download Project Files and Install Dependencies

- [x] **Fork** and **Clone** this repository.
- [x] **CD into the folder** where you cloned the repository.
- [x] Take the steps necessary to create a `package.json` to keep a record of all dependencies
- [x] Add necessary dependencies: `express`, `mongoose` and `nodemon`
- [x] Configure an _npm script_ named _"start"_ that will execute your code using _nodemon_ 

### Design and build endpoints
- [x] A`POST` at `/api/register` which creates a `user`
  - [x] Build user schema and model
  - [x] Hash the password before saving the user to the database
  - [x] Save user with data sent on request body
- [x] A `POST` at `/api/login` which logs in a `user`
  - [x] Compare credentials sent on request body to saved credentials
  - [x] On successful login, create a new session for the user and send back a 'Logged in' message and a cookie that contains the user id
  - [x] If login fails, repond with the correct status code and the message: 'You shall not pass!'
- [x] A `GET` at `/api/users` which gives a logged in user an array of all users in database
  - [x] If the user is not logged in repond with the correct status code and the message: 'You shall not pass!'

### Stretch Problem


dresean:
- Routes
Looking good and organized!
-Models
No complaints!
- Server 
All is well, the abstraction of the db is a but much considering how show and important that is to see in the server. 

- Notes
make more errors specific with error.message from node 
missing mongoose promise = global promise

-Overall
Crushing it as usual! Keep it up!

d-vail:
Can you go over why one would need to set mongoose.Promise to global.Promise. The documentation makes it seem like by default mongoose already allows the use of promises (or at least promise like behavior) and examples have worked without explicitly setting it. I've actually never set it to the global promise. But, if you wanted to use a different promise library, you could then set it with mongoose.Promise: http://mongoosejs.com/docs/promises.html

dresean:
Server -
Love the use of the global middleware from the stretch problem. Your code is becoming easier to read each iteration! Organized perfectly. Very smart coding. 

Models - Looking good!

Middleware - 
beautiful work on this middleware aside from the notes listed below, it is spotless. Works like a charm. 
Routes - All of them work! 

Notes 
- the use of tenaries instead of then & catch is recommended  in the server for checking errors. I would try to stick to the promise based callbacks for better performance and security throughout the server. 
- Although this exercise is small enough to handle multiple instances of the same router object, I would try to take advantage of the chainable methods the router can take to decrease file size and time to locate/debug methods. 
Missing global promise object. 
Missing process.env.port for usablility across platforms. 

Overall
- yet another banger Danny! Keep it up!

dresean:
When connecting mongoose with other libraries, their promise object might not be exactly the same as that libraries promise, that's why we use the global promise object to make sure that that is not the case.

dresean:
Also queries in mongoose are not promises, so when executing queries we can tell mongoose to use promises.

d-vail:
Queries may not be promises but the docs say that asynchronous operations in mongoose will return ES6 promises. I did some digging and found that older repos by Lambda used version 4 of mongoose. Version 4 did not return ES6 promises; asynchronous operations returned mpromises. If you wanted to use native promises you had to set mongoose.Promise to global.Promise. See _Plugging in your own Promises Library_ on http://mongoosejs.com/docs/4.x/docs/promises.html. As of version 5, which we are now using, mongoose uses native promises by default making setting it again unnecessary. See https://github.com/Automattic/mongoose/issues/2917. It is also no longer suggested in the docs. See http://mongoosejs.com/docs/promises.html.

dresean:
Thanks for that information Danny! Looks like you're correct about not truly needing the mongoose.Promise = global.Promise bit! Works like a charm still anyhow :)

PR #185, opened 2018-06-11
Il Ho Kim

ilhoKim:


server.js
```diff
@@ -0,0 +1,112 @@
...
+        maxAge: 1000 * 60 * 60 * 24 // a day
+    },
+    httpOnly: true,
+    secure: false,
+    resave: true,
+    saveUninitialized: false,
+    name: 'noname',
+};
+
+function protected(req, res, next) {
```
etisdew: Looks great.

server.js
```diff
@@ -0,0 +1,112 @@
...
+
+server.get('/', function (req, res) {
+    if (req.session && req.session.username) {
+        res.status(200).json({ message: `Welcome back ${req.session.username}` });
+    } else {
+        res.status(401).json({ message: 'Speak friend and enter' });
+    }
+});
+
+// server.post('/api/register', (req, res) => {
```
etisdew: I love the practice of commenting iterations as it tells a story but this is a 1:1 duplicate haha
Doesn't effect anything just found it odd.

ilhoKim: Those are of day 1 / day 2 comparison note for me.
Commit history was not enough haha

PR #186, opened 2018-06-11
Brandon Hopper

Hopperb:


etisdew:
Amazing performance on this repository, knocked it out in record time and reached to stretch. Didn't have any need for assistance so I'm not sure you need my help here. :100: I'll follow up if I find anything though.

PR #190, opened 2018-06-11
Govi Escudero

jaklundun:


package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "auth-i",
+  "version": "1.0.0",
+  "description": "## Topics",
+  "main": "server.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
enitchals: Remember that you'll also need a `"start": "nodemon"` item in this `scripts` list, for when you want to run your server :)

PR #193, opened 2018-06-11
lonnie davis

ldavis0866:


server.js
```diff
@@ -0,0 +1,129 @@
...
+        maxAge: 1000 * 60 * 60 //an hour
+    },
+    httpOnly: true,
+    secure: false,
+    resave: true,
+    saveUninitialized: false,
+    name: 'noname',
+};
+
+function protected(req, res, next){
```
JacquelynnRohrich: When running your code I found an error and perhaps you may have seen it. This word "protected" is a reserved word so you cannot name functions/variables reserved words. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode. There is some documentation on it. So to fix it name it perhaps restricted instead of protected 😉 guessing it will help a tad! Okie Dokie Thanks Lonnie!

JacquelynnRohrich: as per our conversation in the group zoom, I am thinking this may be an error just within Code sand box however still good practice to know there are words out there reserved specifically for something else 😄

PR #197, opened 2018-06-11
Mason Morrow | CS10

masontmorrow:
## Project Setup

* [x] create a `package.json` file
* [x] `yarn/npm install` to install dependencies and dev dependencies
* [x] configure script key to run nodemon
* [x] `yarn/npm start` to start the API
* [x] use Postman to test the API

### *Schema*

  * [x] create `User.js` model

### *Routes*
1. `registerRouter.js`
  * [x] `POST` to `/api/register` to create a user
  * [x] hash the password before storing in the DB.
2. `loginRouter.js`
  * [x] `POST` to `/api/login ` to let the user log in.
  * [x] respond with a `Logged In` message
  * [x] respond with a cookie of the user's ID.
  * [x] respond with a `You shall not pass!` if login fails.
3. `userRouter.js`
  * [x] `GET` from `/api/users` if user is logged in.
  * [x] respond with a `You shall not pass!` if user is not logged in.
  * [x] use this endpoint to verify that the password is hashed before saving.

## Stretch Problems

  * [x] Write a piece of **global** middleware that ensures a user is logged in when accessing _any_ route prefixed by `/api/restricted/`. For instance, `/api/restricted/something`, `/api/restricted/other`, and `/api/restricted/a` should all be protected by the middleware; only logged in users should be able to access these routes.
  * [ ] Build a React application that implements components to register, login and view a list of users.


server.js
```diff
@@ -0,0 +1,33 @@
+// Modules
+const express = require('express');
+const server = express();
+const mongoose = require('mongoose');
+const helmet = require('helmet');
```
JacquelynnRohrich: Thanks for clarifying where this mystery helmet came from! I will have to let Ryan you picked it up! Nice work thinking outside the box.

PR #198, opened 2018-06-11
Jake Henry

Blast3d:


auth/userRouter.js
```diff
@@ -0,0 +1,18 @@
...
+    const user = new User(req.body);
+    user.save()
+      .then(result => res.status(201).json(result))
+      .catch(err => res.status(500).json({
+        error: err.message
+      }));
+  });
+
+
+module.exports = router;
```
nphillips78: You don't need this file. You've already done this within your server.js. I _think_ you might be trying to create routes for both the client and the server, is that what you're doing?

package.json
```diff
@@ -0,0 +1,26 @@
...
+    "bcrypt": "^2.0.1",
+    "cors": "^2.8.4",
+    "express": "^4.16.3",
+    "helmet": "^3.12.0",
+    "mongoose": "^4.13.14"
+  },
+  "devDependencies": {
+    "nodemon": "^1.17.2"
+  }
+}
```
nphillips78: `yarn add express-session`

server.js
```diff
@@ -0,0 +1,86 @@
...
+// });
+
+const server = express();
+
+server.use(cors());
+server.use(express.json());
+
+server.use('/api/user', userRouter);
+
+const session = require('express-session');
```
nphillips78: you already have this on line 6

server.js
```diff
@@ -0,0 +1,86 @@
...
+
+server.use(
+  session({
+    secret: 'nobody toss a dwarf!',
+    cookie: {
+      maxAge: 1 * 24 * 60 * 60 * 1000
+    },
+    httpOnly: true,
+    secure: true,
```
nphillips78: This isn't what was breaking your code, but remember that this needs to be false in development and change to true in production.

server.js
```diff
@@ -0,0 +1,86 @@
...
+// mongoose.connect('mongodb://localhost/cs10').then(() => {
+//   console.log('\n*** Connected to database ***\n');
+// });
+
+const server = express();
+
+server.use(cors());
+server.use(express.json());
+
+server.use('/api/user', userRouter);
```
nphillips78: ^ don't need this line at all

PR #199, opened 2018-06-11
Braden Walker

bradencodes:


package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "auth-i",
+  "version": "1.0.0",
+  "main": "index.js",
+  "scripts": {
+    "start": "nodemon server.js"
```
enitchals: This can just be `"nodemon"` if you specify `server.js` as your value for `"main"` (line 4)

users/userModel.js
```diff
@@ -0,0 +1,30 @@
+const mongoose = require('mongoose');
+const bcrypt = require('bcrypt');
+const ObjectId = mongoose.Schema.Types.ObjectId;
```
enitchals: I don't think you'll need this declaration for this project (I don't think we'll be modeling relationships between documents here), but good work remembering it. Better to have it and not need it than to need it and not have it! 😁

PR #201, opened 2018-06-11
Herbert Hodgson

trustthedata:


server.js
```diff
@@ -0,0 +1,19 @@
...
+mongoose.connect("mongodb://localhost/LoginDB").then(() => {
+  console.log("Connected to the database");
+});
+
+const registerRouter = require("./registerRouter");
+
+const server = express();
+
+server.use(express.json());
+server.use("/api", registerRouter);
```
enitchals: Remember that using just `/api` here will prevent you from writing other Routers that use `/api/...` Have you considered something like `server.use("/register", registerRouter);`?

registerRouter.js
```diff
@@ -0,0 +1,13 @@
+const express = require("express");
+const User = require("./User");
+const router = express.Router();
+
+router.route("/register").post((req, res) => {
```
enitchals: If you change the path for registerRouter to `/register/` (as I mentioned in my comment on server.js), you could change this to just `/` and be able to register by posting to `localhost:5000/register/`

There's nothing _wrong_ with the way you did it, but it's a good idea to think about logical path names so you have some practice when you start on a larger project 😄

enitchals: Hey Herbert -- just realized these paths were specified in the readme, so disregard these! Sorry about that.

PR #203, opened 2018-06-11
Zach Campbell

wajnurfes:
- [x] Hashing
- [x] Session
- [x] Authentication
- [x] Cookies


auth/userModel.js
```diff
@@ -0,0 +1,40 @@
...
+  username: {
+    type: String,
+    unique: true,
+    required: true,
+    lowercase: true
+  },
+  password: {
+    type: String,
+    required: true,
+    minlength: 4
```
nphillips78: In the real world we're definitely going to want a longer password, minlength should _at least_ be 10.

auth/userModel.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+});
+
+userSchema.pre('save', function(next) {
+  console.log('pre save hook');
+  bcrypt.hash(this.password, 16, (err, hash) => {
```
nphillips78: Good

auth/userModel.js
```diff
@@ -0,0 +1,40 @@
...
+  });
+});
+
+//from https://www.mongodb.com/blog/post/password-authentication-with-mongoose-part-1
+userSchema.methods.comparePassword = function(candidatePassword, cb) {
+    bcrypt.compare(candidatePassword, this.password, function(err, isMatch) {
+        if (err) return cb(err);
+        cb(null, isMatch);
+    });
+};
```
nphillips78: I see that this comparePassword code is from that blog post, but I have 2 questions.
1 - Why did you choose to write it this way? 
2 - Why do you feel this is better than the way we've been doing it?

server.js
```diff
@@ -0,0 +1,100 @@
...
+});
+
+const server = express();
+
+
+server.use(express.json());
+
+const protected = (req, res, next) => {
+  if (req.session && req.session.username) next();
+  else res.status(401).json({ message: "Forbidden!"})
```
nphillips78: Good

server.js
```diff
@@ -0,0 +1,100 @@
...
+	res.json({ viewCount: session.viewCount })
+})
+
+server.post('/api/login', (req,res) => {
+  const { username, password } = req.body
+  User.findOne({ username }, (err, user) => {
+    if (err) res.status(500).json({ err })
+    else user.comparePassword(password, (err, auth) => {
+      if (err) res.status(500).json({ err });
+      else if (auth) {
```
nphillips78: Where did auth come from? I couldn't get this post request to work in Postman, and auth is the thing I can't figure out. Am I blind?

auth/userModel.js
wajnurfes: It's not necessarily that I feel it's *better*, but I do like callback functions.  And the logic of "if the password matches, then do the callback function" is very comfortable to me.  The other way you're referring to is just returning a true or false, right?  So this way basically just saves me the trouble of having to do an if (comparePassword(pass)) {dostuff} and lets me instead comparePassword(pass, dostuff).

server.js
wajnurfes: auth comes from isMatch in my comparePassword method, when I do cb(null, isMatch).  isMatch, in turn, comes from the arguments given by bcrypt.compare to its callback function.

It's working fine for me... weird.  I have uncommitted changes, but nothing that should affect that route I don't think.  I'll revert and give it another whirl ...

Yeah, I get a "logged in successfully" message, and then my get requests on /api/users work.

auth/userModel.js
nphillips78: Aha. Makes sense, thank you :smile:

server.js
nphillips78: It's probably something on my end, thank you for confirming that it works.

PR #211, opened 2018-06-11
Edward Gonzalez - auth-i

eddygonzalez9708:


dresean:
auth - i 
- Authenticate
This is a weird name for a folder that contains schema and a controller, for this assignment it works but I wouldn’t go to town on that naming convention in the future for bigger files. 

-model
excellent work, nothing to complain about!

-server
The grouping of similar methods is very nice, nothing going on that’s majorly bad! Looks good!

- Notes 
missing catch and precise error.message properties from some functions. 
missing global promise = mongoose promise. 

-Overall
Killing it per usual. Keep trying to hack at the login part if you can I know you can get it!

PR #215, opened 2018-06-11
Thuy Pham

iamthuypham:
* [x] create new account POST
* [x] log a user in POST
* [x] get user after log in GET


etisdew:
Really glad to see you have a blocker for once yesterday. :+1:

PR #219, opened 2018-06-14
Initial Pull Request

NewbieWanKenobi:
Jesus Christ walks into a bar, hands the innkeeper three nails, and says, "Can you put me up for the night?"

### DIFF
--- a/data/database.js
+++ b/data/database.js
@@ -1,0 +1,7 @@
+const mongoose = require('mongoose');
+
+module.exports = {
+    connectTo: function (database = 'sandbox', host = 'localhost') {
+        return mongoose.connect(`mongodb://${host}/${database}`);
+    }
+}
--- a/server.js
+++ b/server.js
@@ -24,3 +24,5 @@ database.connectTo('AuthMini')
   server.listen(port, () => {
       console.log(`Server up on port ${port}`);
   })
+
+  
--- a/server.js
+++ b/server.js
@@ -7,6 +7,7 @@ const server = express();
 const database = require('./data/database');
 
 const routes = require('./user/routes');
+const user = require('./user/User');
 
 database.connectTo('AuthMini')
   .then(() => {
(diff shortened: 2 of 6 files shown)


data/database.js
```diff
@@ -0,0 +1,8 @@
+const mongoose = require('mongoose');
+// hey let's make some changes in here 
+
+module.exports = {
+    connectTo: function (database = 'sandbox', host = 'localhost') {
```
lefrenk: Very cool! I like the use of default parameters and string interpolation. 👍

server.js
```diff
@@ -0,0 +1,27 @@
+const mongoose = require('mongoose');
```
lefrenk: We don't need to import `mongoose` here since we do all the Mongoose stuff inside `./data/database.js`

server.js
```diff
@@ -0,0 +1,27 @@
...
+const express = require('express');
+const helmet = require('helmet');
+
+const server = express();
+
+const database = require('./data/database');
+
+const routes = require('./user/routes');
+const user = require('./user/User');
```
lefrenk: We don't need to import the User schema here either. It's only being used inside `./user/routes.js`

user/routes.js
```diff
@@ -0,0 +1,49 @@
...
+      .then(user => {
+          res.status(201).json(user);
+      })
+      .catch(err => {
+          res.status(500).json(err);
+      })
+})
+
+router.put('/login', (req, res) => {
+    if (!req.body.username || !req.body.password) {
```
lefrenk: If we move this if statement below `const { username, password } = req.body` we can make it more concise:

```js
const { username, password } = req.body

if (!username || !password) return res.status(422).json({ error: 'Please provide a valid username and password.' })
```

Also, don't forget to include `return` so the function execution ends, and the rest of your route's code is ignored.

lefrenk:
A SQL query goes into a bar, walks up to two tables and asks, "Can I join you?"

Nice work on the assignment, Sean! Would like to see you implement sessions though. :)

PR #225, opened 2018-06-18
Christina Kopecky

ckopecky:
* [x] **Fork** and **Clone** this repository.
* [x] **CD into the folder** where you cloned the repository.
* [x] Take the steps necessary to create a `package.json` to keep a record of all dependencies.
* [x] Configure an _npm script_ named _"start"_ that will execute your code using _nodemon_ so that the server restarts on changes. Make _nodemon_ be a development time dependency only, it shouldn't be deployed to production.
Design and build a set of endpoints listed below.
* [x] | POST   | /api/register | 
Creates a `user` using the information sent inside the `body` of the request.     
* [x] Use `mongoose middleware` to implement password hashing.
* [x] | POST   | /api/login    | 
Use the credentials sent inside the `body` to login the user. On successful login, create a new session for the user and send back a 'Logged in' message and a cookie that contains the user id. If login fails, repond with the correct status code and the message: 'You shall not pass!' |
* [x] Use `local express middleware` and `mongoose model methods` to implement password verification.
* [x] | GET    | /api/users    | 
If the user is logged in, respond with an array of all the users contained in the database. If the user is not logged in repond with the correct status code and the message: 'You shall not pass!'. Use this endpoint to verify that the password is hashed before it is saved.            |
* [x] **Use _Postman_ to test the API as you work through the exercises.**

###Stretch Problems
* [x] Write a piece of **global** middleware that ensures a user is logged in when accessing _any_ route prefixed by `/api/restricted/`. For instance, `/api/restricted/something`, `/api/restricted/other`, and `/api/restricted/a` should all be protected by the middleware; only logged in users should be able to access these routes.
* [x] Build a React application that implements components to register, login and view a list of users.


UserModel.js
```diff
@@ -0,0 +1,37 @@
...
+    },
+    password: {
+        type: String,
+        minlength: 6,
+        required: true
+    }
+});
+
+//hash functionality using bcrypt
+UserSchema.pre('save', function(next) {
```
designerexpert: Solid use of Schema for Validation and Encryption with pre save hooks and methods.

client/authone/src/UserList.js
```diff
@@ -0,0 +1,16 @@
+import React, { Component } from 'react';
+
+class UserList extends Component {
+
+    //axios request on CWM to obtain userList from database
+    
+    render() {
+        return (
+            <div>
+                I'm a UserList
```
designerexpert: beware of quotes here!

server.js
```diff
@@ -0,0 +1,116 @@
...
+server.get('/', (req, res) => {
+    console.log(req.session);
+    if(req.session && req.session.username) {
+        res.status(200).json({message: `welcome back ${req.session.username}`});
+    }
+    else {
+        res.status(401).json({message: "Login to enter"})
+    }});
+
+server.get('/api/users', restricted, (req, res) => {
```
designerexpert: I like the use of local middle-ware here...i wonder if the same would work with a router implementation and apply it to all its nested routes...

PR #227, opened 2018-06-18
Tramane Hall

tramanedhall:
- [x] **Fork** and **Clone** this repository.
- [x] **CD into the folder** where you cloned the repository.
- [x] Take the steps necessary to create a `package.json` to keep a record of all dependencies.
- [x] Configure an _npm script_ named _"start"_ that will execute your code using _nodemon_ so that the server restarts on changes. Make _nodemon_ be a development time dependency only, it shouldn't be deployed to production.

**Design and build a set of endpoints listed below.**
- [x]  POST | /api/register | Creates a `user` using the information sent inside the `body` of the request.     
- [x] Use `mongoose middleware` to implement password hashing.
- [x]  POST | /api/login | Use the credentials sent inside the `body` to login the user. On successful login, create a new session for the user and send back a 'Logged in' message and a cookie that contains the user id. If login fails, repond with the correct status code and the message: 'You shall not pass!' 
- [x] Use `local express middleware` and `mongoose model methods` to implement password verification.
- [x] GET | /api/users | If the user is logged in, respond with an array of all the users contained in the database. If the user is not logged in repond with the correct status code and the message: 'You shall not pass!'. Use this endpoint to verify that the password is hashed before it is saved.  
- [x] **Use _Postman_ to test the API as you work through the exercises.**

###Stretch Problems
- [ ] Write a piece of **global** middleware that ensures a user is logged in when accessing _any_ route prefixed by `/api/restricted/`. For instance, `/api/restricted/something`, `/api/restricted/other`, and `/api/restricted/a` should all be protected by the middleware; only logged in users should be able to access these routes.


server.js
```diff
@@ -0,0 +1,115 @@
...
+server.post('/api/register', (req, res) => {
+    User
+        .create(req.body)
+        .then(user => {
+            res.status(201).json(user);
+        })
+
+        .catch(error => {
+            if(error.code === 11000) {
+                res.status(500).json({error: "Username already taken!"})
```
designerexpert: I like that you went on the journey to identify the MongoDB errors for duplicate unique fields and implemented it into the error checking portion of the registration.

designerexpert:
# Comments
* Good job implementing MVP 👍 
* Hard work pays off in the end, it just takes a lot of contributing as a group and problem solving.
* Remember if you get a chance implement a ReactClient on one of the Auth projects to see the full circle between Front end and Back end. This will help during project week.

PR #228, opened 2018-06-18
kelly manahan

kellymanahans:


server.js
```diff
@@ -0,0 +1,99 @@
...
+    const { username, password } = req.body;
+    User.findOne({ username })
+    .then(user => {
+        if(user) {
+            user
+            .validatePassword(password)
+            .then(passwordsMatch => {
+                if(passwordsMatch) {
+                    res.status(200).json ({ message: 'You got it'});
+                    req.session.username = user.username;
```
designerexpert: this line needs to happen before line 70. Swap them out.

PR #230, opened 2018-06-18
Bee Choeung

Beeness:


designerexpert:
# Comments
* I like that you are committing often and using meaningful commit messages 👍 
* Looks like MVP was achieved in time, and you were able to understand the concepts well. Good Job!
* As always I wish you get some time in the future to at least in one of the Auth projects attempt a front end so you can see it come to full circle font and back end.
* Keep up the hard work, it pays off in the end!

PR #231, opened 2018-06-18
april martinez

april7229:


designerexpert:
#  Comments
* I know you are working efficiently but those commit messages are not very meaningful. Please remember to write a message saying: What was actually accomplished up until that point of the commit, and maybe what is still left to do.

PR #232, opened 2018-06-19
Usman Javed

ujaved931:
## Auth-i

* [x]  Fork and Clone this repository.
* [x] CD into the folder where you cloned the repository.
* [x]  Take the steps necessary to create a package.json to keep a record of all dependencies.
* [x]  Configure an npm script named "start" that will execute your code using nodemon so that the server 
 restarts on changes. Make nodemon be a development time dependency only, it shouldn't be deployed to production.
* [x]  Design and build a set of endpoints listed below.
* [x]  POST | /api/register | Creates a user using the information sent inside the body of the request.
* [x]  Use mongoose middleware to implement password hashing.
* [x]  POST | /api/login | Use the credentials sent inside the body to login the user. On successful login, create a new session for the user and send back a 'Logged in' message and a cookie that contains the user id. If login fails, repond with the correct status code and the message: 'You shall not pass!'
* [x]  Use local express middleware and mongoose model methods to implement password verification.
* [x]  GET | /api/users | If the user is logged in, respond with an array of all the users contained in the database. If the user is not logged in repond with the correct status code and the message: 'You shall not pass!'. Use this endpoint to verify that the password is hashed before it is saved.
* [x]  Use Postman to test the API as you work through the exercises.


server.js
```diff
@@ -0,0 +1,100 @@
...
+
+server.post('/api/login', (req, res) => {
+    const { username, password } = req.body;
+    User.findOne({ username })
+        .then(user => {
+            if (user) {
+                user.validatePassword(password)
+                    .then(result => {
+                        if (result) {
+                            req.session.username = user.username
```
designerexpert: this line should move into the code block of line 80 above the res.status....

server.js
```diff
@@ -0,0 +1,100 @@
...
+                            if (result) {
+                                res.status(200).json({ msg: Authenticated })
+                                result.validatePassword(password)
```
designerexpert: validatePassword is not a function of the result at this point, which ends up being true or false. Line 81 is unnecessary, and does not get executed due to res.status being an implied return.

designerexpert:
# Comments
* MVP Achieved: Good Job! 👍 
* After a few small changes in the code your project will be golden.
* Keep trying participating you will see how it pays off in the end. It is surprising what we can learn from others when we engage in the creative process and problem solving as a collective.
* If you have time attempt a ReactAppClient on one of the Auth projects to see the full circle between Front and Back end.

PR #239, opened 2018-08-06
William DiFulvio [auth-i]

Wdifulvio523:
Initial PR. Set up server and added needed dependencies


metten0:
- [ ] Appropriate number of commits
- [ ] Appropriate commit messages
Part one, due Monday: 
- [x] Use Node.js, 
- [x] Express and 
- [x] Knex 
- [x] build an API that provides Register and Login functionality using SQLite to store User information. Make sure the password is not stored as plain text.

Part two, due Tuesday: 
- [x] Use sessions and 
- [x] cookies to keep a record of logged in users across requests.
Stretch Problem

- [ ] Write a piece of global middleware that ensures a user is logged in when accessing any route prefixed by /api/restricted/. For instance, /api/restricted/something, /api/restricted/other, and /api/restricted/ashould all be protected by the middleware; only logged in users should be able to access these routes.

MVP! Nice work! Please let me know if you have questions.

Thanks,

Jason

Wdifulvio523:
Thanks! I was wondering how many commits would be appropriate for this project? And how could I make my commit messages more appropriate? I noticed they're not checked off, so I want to continue to improve on this.

metten0:
William,

Sorry, that's on me. I always suggest that students push a commit whenever they have an _ah _ha__ moment, or when they complete a task. You have done that. I also suggest that the commit messages are written as though you are giving report to the next engineer coming on to work on your code. You are among the best in our section at this. In short, I just forgot to scroll all the way up on my form as I did the review. Most humble apologies. In my defense, I have a radish bandit on my hands, and I am very distracted.

Thanks,

Jason

PR #246, opened 2018-08-06
Ari Mercado - auth-i

ari7946:


metten0:
- [x] Appropriate number of commits
- [x] Appropriate commit messages
Part one, due Monday: 
- [x] Use Node.js, 
- [x] Express and 
- [x] Knex 
- [x] build an API that provides Register and Login functionality using SQLite to store User information. Make sure the password is not stored as plain text.

Part two, due Tuesday: 
- [x] Use sessions and 
- [x] cookies to keep a record of logged in users across requests.
Stretch Problem

- [x] Write a piece of global middleware that ensures a user is logged in when accessing any route prefixed by /api/restricted/. For instance, /api/restricted/something, /api/restricted/other, and /api/restricted/ashould all be protected by the middleware; only logged in users should be able to access these routes.
- [x] Build a React application that implements components to register, login and view a list of users. Gotta keep sharpening your React skills.



MVP + stretch! Nice work. User list still doesn't display on frontend, but all other functionality is there. Thanks for your hard work. Please let me know if you have questions.

Thanks,

Jason

PR #252, opened 2018-08-06
Blake Fletcher - auth-i

blkfltchr:


blkfltchr:
Installed and configured `express-session`
Able to store session data at `/getname`, able to read session data at `/setname`
Saved `user.email` as part of the session on` /register` and `/login` and customized welcome messages
Created `protected` function, added it to `GET /users`
Created `/logout` using `destroy`

PR #289, opened 2018-09-10
Das G. MA - AUTH-I

DasGMA:


kelfro:
I will give a more detailed review later but I just wanted to say that it looks good at first glance, and I really like the regular commits and commit messages! Thanks for committing as you go along.

kelfro:
Nice, using 14 for the hash! Everything works and is easy to follow. Try working on one or both of the stretch goals - the react app may be good just to keep those skills sharp. Great job.

kelfro:
Code is easy to follow and well-written. Register and login work well. However, I was able to access users without being logged in. Did you get the same result? In the README it says that if the user is not logged in there should be an error message when they attempt to GET users.
Otherwise, everything is great! Just that fix would make it perfect. Nice work :)

kelfro:
Ah, my bad - I can see where you have the protected function and that should work if activated for production. Excellent progress - the react app looks great. I like the table format.
I rate this a 3 - exceeding expectations :)

PR #291, opened 2018-09-10
Shannon Atkinson / auth-i

sca4cs:
- [x]  POST    /api/register  Creates a `user` using the information sent inside the `body` of the request. **Hash the password** before saving the user to the database.                                                                                                                                                 
- [x]  POST    /api/login     Use the credentials sent inside the `body` to authenticate the user. On successful login, create a new session for the user and send back a 'Logged in' message and a cookie that contains the user id. If login fails, respond with the correct status code and the message: 'You shall not pass!' 
- [x]  GET     /api/users     If the user is logged in, respond with an array of all the users contained in the database. If the user is not logged in repond with the correct status code and the message: 'You shall not pass!'. Use this endpoint to verify that the password is hashed before it is saved.


index.js
```diff
@@ -0,0 +1,109 @@
...
+    .insert(creds)
+    .then(ids => {
+        const id = ids[0];
+        res.status(201).json(id);
+    })
+    .catch(err => res.status(500).send(err));
+});
+
+server.post('/api/login', (req, res) => {
+    if (req.session.username) {
```
wvandolah: This is clever.

index.js
```diff
@@ -0,0 +1,109 @@
...
+    db('users')
+    .where({username: username})
+    .select('id', 'username', 'role')
+    .then(user => {
+        res.json(user);
+    })
+    .catch(err => res.send(err));
+});
+
+server.get('/api/admins', protected, (req, res) => {
```
wvandolah: good job adding roles. looks like you have a strong understanding of the material covered.

PR #294, opened 2018-09-10
Sean Kennedy

skennedy93:


index.js
```diff
@@ -0,0 +1,117 @@
...
+      })
+      .catch(err=>res.status(500).send(err)); 
+})
+server.post("/api/login", (req, res)=>{
+  const creds = req.body; 
+  db("users")
+  .where({username: creds.username})
+  .first()
+  .then(user=>{
+      if(user && bcrypt.compareSync(creds.password, user.password)){
```
adfaris: we a check here at line 64, once the username is valid, then we check our password  against the hashed password.
Good Job finishing the MVP.

PR #296, opened 2018-09-10
John O'Rourke's Auth. Project

johnoro:
# Authentication Project

### Checklist

**Day One**:
   - [x] Register (post)
   - [x] Login (post)
   - [x] Users (get)

**Day Two**:
   - [x] Login sessions
   - [x] Users authorization

---

## Topics

- Authentication.
- Express Middleware.
- Password Hashing.

## Assignment

## Instructions

### Download Project Files

- Do your magic!

### Implement Requirements

#### Endpoints

## Stretch Problem


honda0306:
Well done hitting MVP early! Try building the React front end if you can, just to keep up on it.

PR #299, opened 2018-09-10
Jordan Massingill

jordan-massingill:
Initial PR
Added dependencies and package.json


fake_index.js
```diff
@@ -0,0 +1,166 @@
...
+server.get('/setname', (req, res) => {
+  req.session.name = 'Frodo';
+  res.send('got it');
+});
+
+server.get('/greet', (req, res) => {
+  const name = req.session.username;
+  res.send(`hello ${name}`);
+});
+
```
honda0306: I can't remember if we've talked about how to conceal these before, but remember that it isn't safe to push any sort of token/hashing where the public can see. Let me know if you need help concealing them.

honda0306: Referring to lines 119 && 120.

honda0306:
Overall, your repo looks great! There are just a couple of `good practice` things that we can work on.

fake_index.js
jordan-massingill: This was a go-by I was using, it was Luis's code example from lecture. I meant to delete the file before doing any push's, but it seems as though I didn't!

PR #301, opened 2018-09-10
Francis Tse - Authentication Project

francistse23:
**Part one, due Monday: Use Node.js, Express and Knex to build an API that provides Register and Login functionality using SQLite to store User information. Make sure the password is not stored as plain text.**

- Functionality
- [x] Register
- [x] Login
- Endpoints
- [x] POST create user
- [x] POST authentication
- [x] GET users

**Part two, due Tuesday: Use sessions and cookies to keep a record of logged in users across requests.**
- [x] sessions
- [x] cookies


authentication/src/components/Register.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import axios from 'axios';
+
+const URL = 'http://localhost:8000/api'
```
wvandolah: Might want to consider looking into using a .env file to store environment variables.

PR #304, opened 2018-09-10
Jonathan Jamel Holloway

codejoncode:
### **Implement Requirements**

Use Postman to test the API as you work through the exercises.

- [x] /api/register   Creates a user using the information sent inside the body of the request. Hash the password before saving the user to the database.
- [x] /api/login  Use the credentials sent inside the body to authenticate the user. On successful login, create a new session for the user and send back a 'Logged in' message and a cookie that contains the user id. If login fails, respond with the correct status code and the message: 'You shall not pass!'
- [x] /api/users   If the user is logged in, respond with an array of all the users contained in the database. If the user is not logged in repond with the correct status code and the message: 'You shall not pass!'. Use this endpoint to verify that the password is hashed before it is saved.

**### Day 2  adding in session cookies**

- [x] logout route
- [x] check whether user is logged in middleware
- [x] add  express-session
- [x] add connect-session-knex
- [x] create session 
- [x] create KnexSessionStore 
- [x] use newly added functions 

### **Stretch Problem**
- [x] Write a piece of global middleware that ensures a user is logged in when accessing any route prefixed by /api/restricted/. For instance, /api/restricted/something, /api/restricted/other, and /api/restricted/a should all be protected by the middleware; only logged in users should be able to access these routes.


kelfro:
This is terrific. Love the frequent and descriptive commits, and your code looks clean and well organized. Looking forward to seeing the react app! It's good to keep those skills fresh.

One thing - could you change the title of your PR to your name?

kelfro:
Excellent work, love all the commits and comments. Everything tests correctly in Postman, and I'm pleased to see the React App taking shape. It does not seem to work 100% yet but login and register pages display, which is terrific.
This is a 3 - exceeding expectations. Well done!

PR #305, opened 2018-09-10
Grant Reighard

grantreighard:


index.js
```diff
@@ -29,22 +65,42 @@ server.post('/api/register', (req, res) => {
...
 server.post('/api/login', (req, res) => {
+    // grab creds
     const creds = req.body;
 
+    // find the user
     db('users')
         .where({username: creds.username})
         .first()
         .then(user => {
+            // check creds
```
adfaris: We put our checks on line 76 because one we have a valid username, we can then check for the passoword against the hashed password

index.js
```diff
@@ -0,0 +1,130 @@
...
+                res.send('error logging out');
+            } else {
+                res.send('good bye')
+            }
+        });
+    }
+});
+
+// protect this route, only authenticated users should see it
+server.get('/api/users', protected, (req, res) => {
```
adfaris: Good implantation of protected route

PR #306, opened 2018-09-10
Charlie Winslow - auth-i

cwinslow22:
- [x] Part one, due Monday: Use Node.js, Express and Knex to build an API that provides Register and Login functionality using SQLite to store User information. Make sure the password is not stored as plain text.

- [x] Part two, due Tuesday: Use sessions and cookies to keep a record of logged in users across requests.

**Endpoints**

- [x] POST | /api/register | 
Creates a user using the information sent inside the body of the request. Hash the passwordbefore saving the user to the database.

- [x] POST | /api/login | 
Use the credentials sent inside the body to authenticate the user. On successful login, create a new session for the user and send back a 'Logged in' message and a cookie that contains the user id. If login fails, respond with the correct status code and the message: 'You shall not pass!'

- [x] GET | /api/users | 
If the user is logged in, respond with an array of all the users contained in the database. If the user is not logged in repond with the correct status code and the message: 'You shall not pass!'. Use this endpoint to verify that the password is hashed before it is saved.


index.js
```diff
@@ -0,0 +1,110 @@
...
+  db('users')
+    .select('id', 'username', 'password')
+    .then(users => {
+      res.json(users);
+    })
+    .catch(err => res.send(err));
+});
+
+// logout
+server.get('/api/logout', (req, res) => {
```
wvandolah: Very nice, going beyond MVP.

PR #307, opened 2018-09-10
Kyle Meltzer – auth-i

meltzerkyle:
- [x] npm setup
-----
design and build endpoints:
- [x] POST register
- [x] POST login
- [x] GET users

stretch:
- [x] Write a piece of global middleware that ensures a user is logged in when accessing any route prefixed by /api/restricted/. For instance, /api/restricted/something, /api/restricted/other, and /api/restricted/a should all be protected by the middleware; only logged in users should be able to access these routes.


kelfro:
Good job so far! Would like to see more frequent commits (and don't forget to check off your checklist as you go). Looking forward to seeing your session/cookies.

kelfro:
Sessions and cookies test correctly, code looks great - well organized and readable. I see you started on the react app but so far nothing is up - it's good that at least you got it set up! Maybe later sometime you can continue working on this.
This is a 3 - exceeding expectations. Well done!

PR #310, opened 2018-09-10
Steven L

sliw:
- [ ] registration
- [ ] login


index.js
```diff
@@ -0,0 +1,111 @@
...
+      res.status(201).json(users)
+    })
+    .catch(error => res.status(500).json(error))
+})
+
+const port = 3300
+
+server.listen(port, () => {
+  console.log(`server is listening on port ${port}`)
+})
```
adfaris: Good job finishing the MVP, as we talked about yesterday, you should practice adding the front-end to your server. It is a good practice.

PR #314, opened 2018-09-10
Will Kwon-auth-i

wtkwon:


index.js
```diff
@@ -0,0 +1,67 @@
...
+const express=require('express');
+const session=require('express-session');
+const KnexSessionStore=require('connect-session-knex')(session);
+const cors=require('cors');
+const server=express();
+const knex=require('knex');
+const knexConfig=require('./knexfile');
+const db=knex(knexConfig.development);
+
+const sessionConfig={
```
adfaris: Good job setting up Session configuration

index.js
```diff
@@ -0,0 +1,67 @@
...
+        .then(ids=>res.status(201).json(ids[0]))
+        .catch(err=>res.status(500).send(err));
+})
+server.post('/api/login',(req,res)=>{
+    const creds=req.body;
+    db('users')
+        .where({username:creds.username})
+        .first()
+        .then(user=>{
+            if (user && bcrypt.compareSync(creds.password,user.password)){
```
adfaris: Good placement of the password check.

PR #316, opened 2018-09-10
Katherine GIlligan

katiagilligan888:


client/src/App.js
```diff
@@ -0,0 +1,15 @@
...
+class App extends Component {
+  render() {
+    return (
+      <div className="App">
+        <RegisterPage />
+      </div>
+    );
+  }
+}
+
```
adfaris: Good job working on the stretch, setting the register page component. You should also work on the login, so that the logged user can see the users in the database. 
Overall great job.

PR #318, opened 2018-09-11
david pok - auth i

david-pok:
Implement Requirements

 - [x] Use Postman to test the API as you work through the exercises.

Stretch Problem

 - [ ] Write a piece of global middleware that ensures a user is logged in when accessing any route prefixed by /api/restricted/. For instance, /api/restricted/something, /api/restricted/other, and /api/restricted/a should all be protected by the middleware; only logged in users should be able to access these routes.


db/dbConfig.js
```diff
@@ -0,0 +1,4 @@
+const knex = require("knex");
+const knexConfig = require("../knexfile");
+
+module.exports = knex(knexConfig.developement);
```
wvandolah: Check out your export.  The word development is misspelled. this might causing an error when you start your server

david-pok: OMG THANK YOU! I was up until 1 am trying to figure out what was wrong

routes/userRoutes.js
```diff
@@ -0,0 +1,70 @@
...
+        res.status(200).send(`Welcome ${req.session.username}`);
+      } else {
+        res.status(401).json({ message: "Wrong creds" });
+      }
+    })
+    .catch(err => res.status(500).send(err));
+});
+
+//LOGOUT OF THE SESSION HERE
+server.get('/api/logout', (req, res) => {
```
wvandolah: Good job! But server is not defined on this page. I think you meant to put router.get

PR #322, opened 2018-09-17
Keegan Burkett: Open PR

keegn:
# Authentication Project

## Topics

- Authentication.
- Express Middleware.
- Password Hashing.

## Assignment

## Instructions

### Download Project Files

- Do your magic!

### Implement Requirements

#### Endpoints

## Stretch Problem


database/dbConfig.js
```diff
@@ -0,0 +1,5 @@
+const knex = require('knex');
+
+const knexConfig = require('../knexfile.js');
+
+module.exports = knex(knexConfig.development);
```
Mister-Corn: Slick. 👍

index.js
```diff
@@ -0,0 +1,98 @@
...
+    // find the user
+    db('users').where({username: creds.username}).first().then(user => {
+    // check creds
+    if (user && bcrypt.compareSync(creds.password, user.password)) {
+        req.session.username = user.username;
+        res.status(200).send(`Welcome ${req.session.username}`);
+    } else {
+        res.status(401).json({ message: 'You shall not pass!'});
+    }
+    }).catch(err => res.status(500).send(err))
```

```js
// snippet
.catch(err => {
    console.log('/api/loginPOST error:', err);
});
```

Mister-Corn:
Good job. 👍 You demonstrated understanding of the middleware queue, and code is generally clean and readable. Again, nice work.

PR #325, opened 2018-09-17
JAMES PHILLIPS MBP AUTH-I

CASDFJ:
# Authentication Project

## Topics

- Authentication.
- Express Middleware.
- Password Hashing.

## Assignment

## Instructions

### Download Project Files

- Do your magic!

### Implement Requirements

#### Endpoints

## Stretch Problem


database/dbConfig.js
```diff
@@ -0,0 +1,5 @@
+const knex = require('knex');
+
+const knexConfig = require('../knexfile.js');
+
+module.exports = knex(knexConfig.development);
```
Mister-Corn: This is kinda slick. What inspired you to do this? 😲

index.js
```diff
@@ -0,0 +1,108 @@
...
+server.post("/api/login", (req, res) => {
+  const creds = req.body;
+
+  db("users")
+    .where({ username: creds.username })
+    .first()
+    .then(user => {
+      if (user && bcrypt.compareSync(creds.password, user.password)) {
+        // grab roles from user
+        // req.session.roles = roles
```
Mister-Corn: Good comment for authorization

index.js
```diff
@@ -0,0 +1,108 @@
...
+        res.status(401).json({ Error: "Cannot Authorize" });
+      }
+    })
+    .catch(err => {
+      console.log("Error: ", err);
+      res.status(500).json({ Error: "Login Failed" });
+    });
+});
+//============POST LOGIN ENDPOINT============//
+
```
Mister-Corn: Great work so far. 👍 I'll put in a PR comment once we're in the next project.

index.js
```diff
@@ -1,17 +1,38 @@
...
+const server = express();
+
+const login = require("./Login/login.js");
+const configMiddleware = require("./config/middleware.js");
+
+//configure middleware for the server
+configMiddleware(server);
+
+server.get("/download", (req, res, next) => {
```
Mister-Corn: Still not exactly sure what this endpoint is doing.

index.js
```diff
@@ -35,92 +56,29 @@ const sessionsConfig = {
 //======session configuration====//
 
 server.use(session(sessionsConfig));
-server.use(express.json());
-server.use(cors());
```
Mister-Corn: Hopefully you're getting a better hang of this middleware queue 👍

index.js
```diff
@@ -35,92 +56,29 @@ const sessionsConfig = {
...
+server.use("/api", login);
 
 //========middleware==========//
-function auth(req, res, next) {
-  if (req.session && req.session.username) {
-    next();
-  } else {
-    res.status(401).json({ Error: "You shall not pass!!" });
+function errorHandler(err, req, res, next) {
+  console.error(err);
```
Mister-Corn: This `errorHandler` function is new to me, so I'm legitimately curious. How does it work? When does express call it?

Mister-Corn:
Good work. 👍 Hopefully you're getting a better grasp on the middleware queue. I would like to know how that `errorHandler` middleware works. That looks pretty cool!

index.js
CASDFJ: The errorHandler is just middle ware  for when the get(/download) doesn't work

index.js
CASDFJ: The /download is used to connect the routers with the configuration files and index.js. It was also something referenced by Luis in the videos a while back. Same with the error handler.

PR #328, opened 2018-09-17
Alex Caves

AlexanderCaves:


index.js
```diff
@@ -0,0 +1,51 @@
...
+    const hash = bcrypt.hashSync(creds.password, 10);
+    creds.password = hash;
+
+    db('users')
+        .insert(creds)
+        .then(ids => {
+            const id = ids[0];
+            res.status(201).json(id);
+        })
+        .catch(err => res.status(500).send(err));
```

```js
// snippet
.catch(err => {
    console.log('/api/register POST error:', err);
});
```

knexfile.js
```diff
@@ -0,0 +1,51 @@
...
+    },
+    pool: {
+      min: 2,
+      max: 10
+    },
+    migrations: {
+      tableName: 'knex_migrations'
+    }
+  }
+
```
Mister-Corn: Good work so far. I'll put in a PR comment once this project is done.

Mister-Corn:
Good job on this. 👍 Not much to say, other than you're by the book. You seem to understand the middleware queue, as I don't see extraneous calls to middleware functions or encountered any duplicate functionality.

PR #330, opened 2018-09-17
Richard Verdier - Fork & Clone Completed

rverdi642:


index.js
```diff
@@ -0,0 +1,138 @@
...
+  //replace user password with the hash
+  creds.password = hash;
+  //save the user
+
+  db('users').insert(creds).then(ids => {
+
+    const id = ids[0];
+  res.status(201).json(id);
+
+  }).catch(err => res.status(500).send(err));
```

```js
// snippet
.catch(err => {
    console.log('/api/register POST error:', err);
});
```

index.js
```diff
@@ -0,0 +1,138 @@
...
+  //save the usersreurn 201
+
+});
+
+server.post('/api/login', (req,res) =>{
+  const creds = req.body;
+
+  db('users')
+    .where({username: creds.username })//where always returns an array
+    .first()                            //aska for the first item in the array
```
Mister-Corn: Awesome notes and comments. 👍

index.js
```diff
@@ -0,0 +1,138 @@
...
+  // const name = req.session.name;
+  const name = req.session.username;
+  // res.send(`hello ${req.session.name}`);
+  res.send(`hello ${name}`);
+});
+
+// protect this route, only authenticated users should see it
+server.get('/api/users', (req, res) => {
+  if(req.session && req.session.username) {
+  db('users')
```
Mister-Corn: This is good, but since you have a `protected` middleware function, you could also do something like this:

```js
server.get('/api/users', protected, (req, res) => {
    // blah blah blah
});
```

The `protected` middleware will run first, and call `next()` (_and thus run the next middleware function_) upon successful authentication, or send an error response and end the request/response if unsuccessful.

index.js
```diff
@@ -0,0 +1,138 @@
...
+    })
+    .catch(err => res.send(err));
+  }
+  res.status(401).json({message: "Unauthorized attempt"});
+
+});
+
+server.get('/api/admins', (req, res) => {
+  //only see users, if you have the role of 'admin'
+  if (req.session && req.session === 'admin') {
```
Mister-Corn: I'll have to double-check this, but you might be overwriting the entire `session` object here. You might want to do something like `req.session.role === 'admin'` or something like that.

index.js
```diff
@@ -0,0 +1,138 @@
...
+    })
+    .catch(err => res.send(err));
+  // }
+} else {
+  res.status(403).json({message: "Role violation - No access to resource"});
+}
+  // res.status(401).json({message: "Unauthorized attempt"});
+  // 
+});
+
```
Mister-Corn: Great job so far. 👍 I'll leave a PM comment once this project is done.

Mister-Corn:
Good work on this. 👍 I think I discussed everything I needed to on the comments.

PR #334, opened 2018-09-17
Erica Thompson: MVP Complete Day 2 - In Stretch

EricaThompson:


db/migrations/20180917141421_users_table.js
```diff
@@ -0,0 +1,22 @@
+
+exports.up = function(knex, Promise) {
+  return knex.schema.createTable('users_table', (tbl) => {
+    
+    tbl
+    .increments()
+
+    tbl
+    .string('username', 128)
+    .notNullable()
```
Mister-Corn: I recommend using `.unique()` here to ensure there are no duplicate usernames. I think that may cause problems with login if there are. Otherwise, this is looking nice. 👍

server.js
```diff
@@ -0,0 +1,72 @@
...
+        })
+        .catch(err => {
+            console.log('/api/register POST error:', err);
+            res.status(500).send('Please try again later');
+        });
+});
+
+server.get('/api/users', (req, res) => {
+    db('users_table').select('id', 'username').then(users => {
+        res.status(201).json(users);
```
Mister-Corn: In this endpoint, we are getting an array of all users in the `users_table`. We are not creating a new user, so I feel HTTP status code `200` (_OK/Success_) is more appropriate than `201` (_Created_).

server.js
```diff
@@ -0,0 +1,72 @@
...
+    db('users_table')
+    .where({username: creds.username})
+    .first()
+    .then(user => {
+        if (user && bcrypt.compareSync(creds.password, user.password)){
+            res.status(200).send('Welcome');
+        } else {
+            res.status(401).json({message: 'incorrect combination'});
+        }
+    }).catch(err => {res.status(500).send(err, "Everything failed")});
```
Mister-Corn: Make sure you put a console.log in your catches so you can troubleshoot errors more effectively:

```js
// snippet
.catch(err => {
    console.log('/api/login POST error:', err);
    res.status(500).send("Everything failed");
});
```

server.js
```diff
@@ -0,0 +1,72 @@
...
+})
+
+
+
+
+
+
```
Mister-Corn: Beautiful work so far. 👍 I'll add a PR comment once we're done with this project.

db/migrations/20180917141421_users_table.js
EricaThompson: That's awesome, I tried that with Rachel this morning because she had it, but after trying it both ways, it didn't seem to change the ability to create a user from my tests. I ended up not using the code since I realized I couldn't fully explain if I needed it or not.

server.js
EricaThompson: Ahh, true, thanks

db/migrations/20180917141421_users_table.js
Mister-Corn: As long as there's no duplicates, you shouldn't see any problems. However `.unique()` _enforces_ no duplicates, whereas without it, there's no problems until someone has the same username as someone else. Then your `login` route will end up returning two people and mess that up.

server.js
Mister-Corn: I wrote this in another student's PR:

> I'd recommend console.logging the error in you `catch()` block. The error tends to be shortened in PostMan, and conceptually, it's not the client's job to fix your server, so they don't really need to know the exact error.

db/migrations/20180917141421_users_table.js
Mister-Corn: Or are you saying you tried to add duplicate users and `.unique()` did nothing?

db/migrations/20180917141421_users_table.js
```diff
@@ -12,6 +12,11 @@ exports.up = function(knex, Promise) {
     tbl
     .string('password', 128)
     .notNullable()
+    .unique();
+
+    tbl
+    .string('role', 128)
+    .notNullable();
```
Mister-Corn: Not part of MVP, but a nice starting point to implement authorization in this API.

server.js
```diff
@@ -36,37 +76,108 @@ server.post('/api/register', (req, res) => {
...
+                res.send('good bye');
+            }
+        });
+    }
+});
+
+server.get('/api/users', restricted, (req, res) => {
+    
+    if(req.session && req.session.username){
+
```
Mister-Corn: Notice in line 112 you have the `restricted` middleware listed. Do you see any issue with this and line 114? Let me know!

server.js
```diff
@@ -36,37 +76,108 @@ server.post('/api/register', (req, res) => {
...
+        .select('r.role_name')
+        .where('ur.user_id', userId)
+        .then(roles => {
+            if(roles.includes('admin') || roles.includes('boss')){
+                //have access
+            } else {
+                //bounce
+            }
+        })
+    }
```
Mister-Corn: I think this is a good approach. Another approach is to save the role in `req.session`, like `req.session.role`. Then, you can merely read from the session to continue with authorization.

Mister-Corn:
Good job so far. 👍 I appreciate the little stretch on implementing authorization on your project. I'd like you to review the use of `restricted` I mentioned in one of my comments.

PR #336, opened 2018-09-17
Tim Zulf MBP- MVP

Tulf:
## Assignment

## Instructions

### Download Project Files

- Do your magic!

### Implement Requirements

#### Endpoints

## Stretch Problem


index.js
```diff
@@ -0,0 +1,64 @@
...
+ const credentials = req.body;
+ db("credentials")
+   .where({ username: credentials.username })
+   .first()
+   .then(user => {
+     if (user && bcrypt.compareSync(credentials.password, user.password) ) {
+       res.status(200).send("Hello my friend, nice to see you again");
+     }
+     else {
+       res.status(401).json({ message: "Man you can't even type right " });
```
Mister-Corn: Shaming "butter-finger" users is the best. 😂

seeds/users_info.js
```diff
@@ -0,0 +1,15 @@
...
+  // Deletes ALL existing entries
+  return knex('credentials').del()
+    .then(function () {
+      // Inserts seed entries
+      return knex('credentials').insert([
+        {username: 'bobheper1', password: 'biallzabob'},
+        {username: 'ricardo', password: 'david'},
+        {username: 'Greg', password: 'mankiw'},
+        {username: 'orungatan', password: 'spatialmemory'},
+        {username: 'eric', password: 'elhricson'}
```
Mister-Corn: Are you able to log in with these users? The passwords stored in the server should be hashed, so trying to log-in with these might not work, because you'll be hashing the client input, but the password in the database is in plaintext.

An idea; import `bcrypt` then:
```js
{username: 'bobheper1', password: bcrypt.hashSync('biallzabob', 14)},
```
That might work for seeding.

Tulf: Oh you're right I didn't test these I only tested the ones I passed in. thanks!

index.js
```diff
@@ -0,0 +1,117 @@
...
+const sessionConfig = {
+  name: 'login',
+  secret: 'fly you fools',
+      cookie: {
+        maxAge: 1 * 24 * 60 * 60 * 1000,
+        secure: false, // only set cookies over https. Server will not send back a cookie over http.
+      },
+      resave:false,
+      saveUninitialized:false
+}
```
Mister-Corn: What turned you off from using `KnexSessionStore`?

index.js
```diff
@@ -0,0 +1,117 @@
...
+server.use(session(sessionConfig))
+
+
+//proteced middleware for protected route:
+function protected (req, res, next) {
+  if(req.session && req.session.username){
+    next();
+  }
+  else{
+    res.status(400).json({error: 'Session credentials incorrect for this user' })
```
Mister-Corn: `400` (_Bad Request_) is good. Another possibility is `401` (_Unauthorized_).

seeds/users_info.js
```diff
@@ -0,0 +1,15 @@
...
+  // Deletes ALL existing entries
+  return knex('credentials').del()
+    .then(function () {
+      // Inserts seed entries
+      return knex('credentials').insert([
+        {username: 'bobheper1', password: bcrypt.hashSync('biallzabob',14)},
+        {username: 'ricardo', password: bcrypt.hashSync('david',14)},
+        {username: 'Greg', password: bcrypt.hashSync('mankiw',14)},
+        {username: 'orungatan', password: bcrypt.hashSync('spatialmemory',14)},
+        {username: 'eric', password: bcrypt.hashSync('elhricson',14)}
```
Mister-Corn: This is cool. Let me know if it worked!

Mister-Corn:
Overall, good job. 👍 Let me know why `KnexSessionStore` wasn't jiving with you.

index.js
Tulf: Nothing huge I just didn't see it as being necessary. I guess I understand how the session is stored and cookies are passed. To me it'  more intuitive to just set the session configuration explicitly.

PR #337, opened 2018-09-17
JoCatherine Griggs - Authentication 

jocafrin:
Implement Requirements

- [ ] Use Postman to test the API as you work through the exercises.

#### Endpoints

 - [x] | POST   | /api/register | Creates a `user` using the information sent inside the `body` of the request. **Hash the password** before saving the user to the database.                                                                                                                                                 |
- [x] |  POST   | /api/login    | Use the credentials sent inside the `body` to authenticate the user. On successful login, create a new session for the user and send back a 'Logged in' message and a cookie that contains the user id. If login fails, respond with the correct status code and the message: 'You shall not pass!' |
- [x] | GET    | /api/users    | If the user is logged in, respond with an array of all the users contained in the database. If the user is not logged in repond with the correct status code and the message: 'You shall not pass!'. Use this endpoint to verify that the password is hashed before it is saved.            |


index.js
```diff
@@ -0,0 +1,59 @@
...
+    secret: 'a clock is right twice a day',
+    cookie: {
+      maxAge: 1 * 24 * 60 * 60 * 1000,
+      secure: false,
+    },
+    httpOnly: true,
+    resave: false,
+    saveUninitialized: false,
+  };
+
```
Mister-Corn: Nice getting this set up!

index.js
```diff
@@ -0,0 +1,59 @@
...
+    .first()
+    .then(user => {
+      if (user && bcrypt.compareSync(creds.password, user.password)) {
+        res.status(200).send("Login Successful");
+      } else {
+        res.status(401).json({ message: "You are not authorized." });
+      }
+    })
+    .catch(err => res.status(500).send(err));
+});
```
Mister-Corn: Good work so far. 👍 I'll put a general comment in the PR after we enter the next project for the week.

PR #342, opened 2018-10-15
Huthman King: Auth I

kinghuthman:
@bradiowave


routes/userRoutes.js
```diff
@@ -0,0 +1,77 @@
...
+    creds.password = hash; //overrides, password = hash
+    
+    db("users")
+        .insert(creds) //pass credentials within users table
+        .then(ids => {
+            const id = ids[0]; //array grab the first id
+            req.session.username = user.username;
+            res.status(201).json(id); // sends back new user id
+        })
+        .catch(err => res.status(500).send(err));
```
Bradiowave: Write your own error message to describe why the user could not register.

routes/userRoutes.js
```diff
@@ -0,0 +1,77 @@
...
+});
+
+
+
+router.post("/login", (req, res) => {
+    const creds = req.body;
```
Bradiowave: Specify that you only want the username and password from the request body.

Bradiowave:
## MVP
    - POST to /api/register 
    - POST to /api/login
    - GET to /api/users 

## Stretch

## What Looks Great
- MVP complete! Nice work
- Descriptive commits
- Lots of comments on the register endpoint.
- Great file structure. Routes, database migrations, and seeds are all in separate folders.

## What Could Be Improved
- You should have one commit for every one change
- Comment you login and users endpoints just like you did with register

## What Could Be Added
- Move on to stretch for extra practice!

## Rating (1-3)
**2**

PR #345, opened 2018-10-15
Auth-I (lucasbeemer)

lucasbeemer:
@bradiowave


index.js
```diff
@@ -0,0 +1,131 @@
...
+  credentials.password = hash;
+
+  db('users')
+    .insert(credentials)
+    .then(ids => {
+      const id = ids[0];
+      res.status(201).json({ newUserId: id });
+    })
+    .catch(err => {
+      res.status(500).json(err);
```
Bradiowave: Write your own error message to make the error and how to fix it more obvious.

index.js
```diff
@@ -0,0 +1,131 @@
...
+  const jwtOptions = {
+    expiresIn: '5m',
+  };
+  return jwt.sign(jwtPayload, jwtSecret, jwtOptions);
+}
+
+
+// ========== LOGIN ========= //
+server.post('/login', (req, res) => {
+  const creds = req.body;
```
Bradiowave: It would be best to specify that you want only the username and password from the request body.

Bradiowave:
## MVP
    - POST to /api/register 
    - POST to /api/login
    - GET to /api/users 

## Stretch

## What Looks Great
- MVP complete! Nice work
- Good amount of commits, still need some work on the messages
- Comment headers for each chunk of code look great
- Good file structure. Migrations are in their own folder

## What Could Be Improved
- You should have one commit for every one change.

## What Could Be Added
- Add more comments to explain what each chunk of code does.
- Go on to the stretch for more practice!

## Rating (1-3)
**2**

PR #350, opened 2018-10-15
Auth I

VeraSimon:
@bradiowave


index.js
```diff
@@ -0,0 +1,134 @@
...
+// find() -> [{id: int, username: 'string'}, ..., {id: int, username: 'string'}]
+server.get('/api/restricted/users', (req, res, next) => {
+    usersTable.find()
+        .then((users) => {
+            res.status(200).json(users);
+        })
+        .catch((err) => {
+            next(["h500", err]);
+        });
+});
```
Bradiowave: I like that you included both ways to include authorization middleware.

Bradiowave:
## MVP
    - POST to /api/register 
    - POST to /api/login
        - [x] Use the credentials sent inside the body to authenticate the user. On   
        successful login, create a new session for the user and send back a 'Logged in' message and a 
        cookie that contains the user id. 
    - GET to /api/users 
        - [x] If the user is not logged in repond with the correct status code and the message: 'You shall not  
        pass!'.

## Stretch
- [x] Write a piece of global middleware that ensures a user is logged in when accessing any route prefixed by /api/restricted/.

## What Looks Great
- MVP complete! Nice work
- Stretch also complete, great work Vera!
- Frequent and descriptive commits
- Great file structure! Each component has its own folder. Migrations and database helpers are in their own data folder.
- Error handling middleware!
- Comments both label and explain various parts of the code, like endpoints and middleware.

## What Could Be Improved
- You could move some of the files like `index.js` and `knexfile.js` into their appropriate directories instead of leaving them in the root folder.

## Rating (1-3)
**3**

PR #352, opened 2018-10-15
mvp

agreb17:
@zackhitch


data/migrations/20181015153135_users.js
```diff
@@ -0,0 +1,15 @@
+exports.up = function(knex, Promise) {
```
zackhitch: 👍

index.js
```diff
@@ -0,0 +1,108 @@
...
+
+const sessionConfig = {
+  secret: "dont-steal.my%secret!",
+  name: "someWierdAnimal",
+  httpOnly: true,
+  resave: false,
+  saveUninitialized: false,
+  cookie: {
+    secure: false,
+    maxAge: 1000 * 60 * 1
```
zackhitch: How long is this?

index.js
```diff
@@ -0,0 +1,108 @@
...
+      } else {
+        res.send("See Ya");
+      }
+    });
+  }
+});
+
+server.post("/register", (req, res) => {
+  const creds = req.body;
+  const hash = bcrypt.hashSync(creds.password, 14);
```
zackhitch: For development purposes, its okay, and recommended, to use a lower number of rounds, say something like 7 or 8, so you don't spend too much time waiting around. But then you'd bump it up to at least 14 for production.

index.js
```diff
@@ -0,0 +1,108 @@
...
+
+  db("users")
+    .where({ username: credentials.username })
+    .first()
+    .then(user => {
+      if (user && bcrypt.compareSync(credentials.password, user.password)) {
+        req.session.username = user.username;
+        res.status(200).json({ hello: user.username });
+      } else {
+        res.status(401).json({ message: "you get no access you dirty hacker" });
```
zackhitch: More.... generic error messages should be used.

knexfile.js
```diff
@@ -0,0 +1,14 @@
+// Update with your config settings.
+
+module.exports = {
```
zackhitch: 👍

PR #353, opened 2018-10-15
MVP

PatrickTheCodeGuy:
- [X ] Monday MVP


index.js
```diff
@@ -0,0 +1,103 @@
...
+			if (err) {
+				res.send("You cant leave!");
+			} else {
+				res.send("goodbye!");
+			}
+		});
+	} else {
+		res.send(500).send("No Session To Log Out From");
+	}
+});
```
Bradiowave: Nice inclusion of logout functionality to complete the package.

index.js
```diff
@@ -0,0 +1,103 @@
...
+		.first()
+		.then(user => {
+			if (user && bcrypt.compareSync(creds.password, user.password)) {
+				//Cookie made for session
+				req.session.username = user.username;
+				res.status(201).json({ welcome: user.username });
+			} else {
+				res
+					.status(500)
+					.json({ error: "Wrong Username and/or Password, please try again" });
```
Bradiowave: Make all of your formatting consistent. Line 80 and lines 82-84 should look the same since they are doing very similar things.

Bradiowave:
## MVP
    - POST to /api/register 
    - POST to /api/login
    - GET to /api/users 
        - [x] If the user is not logged in respond with the correct status code and the message: 'You shall not pass!'.

## Stretch

## What Looks Great
- MVP complete! Nice work
- Front end created to connect with api
- Frequent and descriptive commits. Great work on the commits with this project!
- Good file structure. Migrations are in a data folder. Frontend is in it's own directory as well.

## What Could Be Improved
- You can move files like `index.js` and `knexfile.js` into their appropriate folders instead of keeping them on the root. Just make sure to change your file references as accordingly.

## What Could Be Added
- Practice commenting your code, it's a valuable skill.
- The global middleware stretch problem would make for good additional practice.

## Rating (1-3)
**3**

PR #359, opened 2018-10-15
auth-i/Clint Kunz

clintKunz:
@bradiowave


index.js
```diff
@@ -0,0 +1,107 @@
...
+            res.status(200).json(missions);
+        })
+        .catch(err => res.status(500).json(err));
+})
+
+function protected(req, res, next) {
+    if (req.session && req.session.userId) {
+        next();
+    } else {
+        res.status(401).json({ message: 'Not Authorized' });
```
Bradiowave: This is a good start on the stretch, but it isn't global middleware. Global middleware is used when you say `server.use(...)`. For this to work properly as global middleware, it needs to check that the endpoint starts with `/api/restricted`

index.js
```diff
@@ -0,0 +1,107 @@
...
+    const credentials = req.body;
+    const hash = bcrypt.hashSync(credentials.password, 14);
+    credentials.password = hash;
+
+    db('users').insert(credentials).then(ids => {
+        const id = ids[0];
+        req.session.userId = id;
+        res.status(201).json({ newUserId: id });
+    }).catch(err => {
+        res.status(500).json({ err });
```
Bradiowave: Return your own error message explaining why the user could not register.

index.js
```diff
@@ -0,0 +1,107 @@
...
+    });
+})
+
+//Post check user
+server.post('/api/login', (req, res) => {
+    const credentials = req.body;
```
Bradiowave: Specify that you only want username and password from the request body.

Bradiowave:
## MVP
    - POST to /api/register 
    - POST to /api/login
    - GET to /api/users 

## Stretch
- [x] Write a piece of global middleware that ensures a user is logged in when accessing any route prefixed by /api/restricted/.

## What Looks Great
- MVP complete! Nice work
- Stretch work is well done!
- Comments label each endpoint
- Great file structure. Database data has it's own folder. The client has its own folder.

## What Could Be Improved
- Need more commits. You should have one commit for every one change
- The middlware stretch goal needs a little work to make it global middleware

## Rating (1-3)
**3**

PR #367, opened 2018-10-15
Caitlin G's Auth-i

caitlin-g:
@bradiowave


users/usersRoutes.js
```diff
@@ -0,0 +1,104 @@
...
+
+  db("users")
+    .insert(credentials)
+    .then(ids => {
+      const id = ids[0];
+      req.session.username = user.username;
+      res.status(201).json({ newUserId: id });
+    })
+    .catch(err => {
+      res.status(500).json(err);
```
Bradiowave: Send back your own error message saying why they failed to get the users.

users/usersRoutes.js
```diff
@@ -0,0 +1,104 @@
...
+});
+
+//Middleware for protected endpoints
+function protected(req, res, next) {
+  if (req.session && req.session.username) {
+    next();
+  } else {
+    res.status(401).json({
+      message: "Not Authorized"
+    });
```
Bradiowave: This is a good start on the stretch, but it isn't global middleware. Global middleware is used when you say `server.use(...)`. For this to work properly as global middleware, it needs to check that the endpoint starts with `/api/restricted`.

Bradiowave:
## MVP
    - POST to /api/register 
    - POST to /api/login
    - GET to /api/users 

## Stretch
- [x] Write a piece of global middleware that ensures a user is logged in when accessing any route prefixed by /api/restricted/. (Good start, but isn't used globally)

## What Looks Great
- MVP complete! Nice work
- Started on the middleware stretch!
- Frequent and descriptive commits
- Good file structure. Routes are broken into their own file.

## What Could Be Added
- Add comments to each endpoint explaining what it takes from the request, and what it returns as a successful response.

## Rating (1-3)
**3**

PR #370, opened 2018-10-15
auth-i

upsmancsr:
@zackhitch


authHelpers/authRoutes.js
```diff
@@ -0,0 +1,105 @@
...
+  cookie: {
+    secure: false,
+    maxAge: 1000 * 60 * 1
+  },
+};
+
+router.use(session(sessionConfig));
+
+// register a new user
+router.post('/register', isAuthenticated, (req, res) => {
```
zackhitch: When registering a user, they wouldn't be authenticated, so why have `isAuthenticated` middleware be used here?

authHelpers/authRoutes.js
```diff
@@ -0,0 +1,105 @@
...
+            message: `You shall not pass... as you are not authenticated.`,
+          });
+        }
+      })
+      .catch(err => {
+        res.status(500).json(err);
+      });
+});
+
+router.get('/logout', isAuthenticated, (req, res) => {
```
zackhitch: Does the middleware need to be used here? You're simply logging out the user, by removing the session if it exists. There's not need to call that middleware here.

authHelpers/authRoutes.js
```diff
@@ -0,0 +1,105 @@
...
+          res.send('error logging out');
+        } else {
+          res.send('good bye');
+        }
+      });
+    }
+});
+
+// protect this route, only authenticated users should see it
+router.get('/users', isAuthenticated, (req, res) => {
```
zackhitch: This being the protected route, definitely _needs_ to have the `isAuthenticated` middleware.

authHelpers/authRoutes.js
```diff
@@ -0,0 +1,105 @@
+const express = require('express');
```
zackhitch: Your naming convention is a bit odd. `authHelpers` and `authRoutes` don't necessarily describe what's going on. A directory for `routes` and then individual files for the `register`, `login`, `logout`, etc routes, or all of those routes in an `index` file in the `routes` directory would be more clear.

data/dbConfig.js
```diff
@@ -0,0 +1,7 @@
+const knex = require('knex');
```
zackhitch: 👍

data/migrations/20181015142904_users.js
```diff
@@ -0,0 +1,15 @@
+exports.up = function(knex) {
```
zackhitch: Very good.

data/migrations/20181015142904_users.js
```diff
@@ -0,0 +1,15 @@
...
+    return knex.schema.createTable('users', users => {
+      users.increments();
```
zackhitch: Personal preference here, but I like to put whitespace between each field for better readability.

PR #378, opened 2018-10-15
jcassidy

cassidyjamesw:


database/migrations/20181015170302_users.js
```diff
@@ -0,0 +1,15 @@
+exports.up = function(knex) {
+  return knex.schema.createTable("users", users => {
+    users.increments();
+
+    users
+      .string("username", 128)
+      .notNullable()
+      .unique();
```
zackhitch: Just personal preference, but I like whitespace between each field so better aid readability.

index.js
```diff
@@ -0,0 +1,112 @@
...
+// GET //
+server.get("/", (req, res) => {
+  res.send("server functional");
+});
+
+server.get("/api/users", protected, (req, res) => {
+  db("users")
+    .select("id", "username")
+    .then(users => {
+      res.json(users);
```
zackhitch: what happened to `res.status`? You should always include that.

PR #381, opened 2018-10-15
mvp + some stretch

easyra:


data/models/userInfoModel.js
```diff
@@ -0,0 +1,31 @@
...
+ return db(table)
+}
+
+function getByUsername(username){
+    return db(table).where({username}).first()
+}
+
+function getAllUsernames(){
+    return db(table).select('username')
+}
```
Bradiowave: Great database helper functions to make the routing more self-explanatory.

data/routers/userInfoRouters.js
```diff
@@ -0,0 +1,72 @@
...
+  });
+});
+
+router.get("/users", (req, res) => {
+  db.getAllUsernames()
+    .then(usernameList => {
+      if (req.session && req.session.userId) {
+       const usernameArray = [];
+        usernameList.forEach(user => {
+          usernameArray.push(user.username);
```
Bradiowave: This is a great idea, but will you ever add more than one user at a time?

data/routers/userInfoRouters.js
```diff
@@ -0,0 +1,72 @@
...
+        });
+        res.json({usernames: usernameArray });
+
+      } else {
+        res.json({ message: "not allowed" });
+      }
+    })
+    .catch(err => res.json(err));
```
Bradiowave: Return your own error message to explain why the user could not register.

data/routers/userInfoRouters.js
```diff
@@ -0,0 +1,72 @@
...
+});
+
+router.post("/register", (req, res) => {
+  const { username, password } = req.body;
```
Bradiowave: This is really good that you specified what you want from the request body.

data/routers/userInfoRouters.js
```diff
@@ -0,0 +1,72 @@
...
+
+router.get("/users", (req, res) => {
+  db.getAllUsernames()
+    .then(usernameList => {
+      if (req.session && req.session.userId) {
+       const usernameArray = [];
+        usernameList.forEach(user => {
+          usernameArray.push(user.username);
+        });
+        res.json({usernames: usernameArray });
```
Bradiowave: Don't request the users from the database until you know that the user is authenticated. This would be a good opportunity to use middleware to check if the user is authenticated.

Bradiowave:
## MVP
    - POST to /api/register 
    - POST to /api/login
    - GET to /api/users 

## Stretch

## What Looks Great
- MVP complete! Nice work
- Code is very well formatted and easy to read
- Great file structure! I like the separation of models, routes and migrations into their own directories.
- Thorough database helper functions

## What Could Be Improved
- Need more commits. You should have one commit for every one change

## What Could Be Added
- Try the stretch problems for more practice!

## Rating (1-3)
**2**

PR #404, opened 2018-11-26
Auth I

CarloC24:


KevinBrack:
Great work Carlo. Server is clean and works well. I was able to register a new user, log in, and retrieve the list of users only when logged in. Keep this up

PR #408, opened 2018-11-26
Blake Koennecker -- auth-i

koennecker:


index.js
```diff
@@ -0,0 +1,102 @@
...
+
+const db = require("./database/dbConfig.js");
+
+const server = express();
+
+const sessionConfig = {
+  name: "login",
+  secret: "asfjaofuwruq04r3oj;ljg049fjq30j4jlajg40j40tjojasl;kjg",
+  cookie: {
+    maxAge: 1000 * 60 * 10,
```
Velsu: That cookie will be outdated the moment it is created because it's maxAge is too short.

PR #410, opened 2018-11-26
auth-i by Katy Hollobaugh

KHollobaugh:


KevinBrack:
Really great work. Server functions exactly as requested. Authentication, registering, and accessing the data only when logged in worked perfectly.

PR #412, opened 2018-11-26
auth-i By Michael Checo

MichaelCheco:


KevinBrack:
Great job Michael. Server functioned as requested. Authorization, registering, login, and viewing data once logged in worked perfectly.

Only thing I would warn, is that the endpoints were specified in the readme as `GET | /api/users`. While changing this is not a huge deal when you are working on a solo project. If you are working on a team, that depends on those endpoints, you would need to make sure you follow the specs provided to you.

PR #413, opened 2018-11-26
auth-i - Pedro Montesinos

pedrolmr:


KevinBrack:
For the most part the server works as requested. I was able to register a new user, login, and get the users from the database. Unfortunately I did not need to be logged in to get access to the users data. You need to create the middle ware to block that data if someone is not logged in.

PR #415, opened 2018-11-26
auth-i- Brian Gregorius

brianmgre:


database/migrations/20181126150111_users.js
```diff
@@ -0,0 +1,19 @@
+
+exports.up = function (knex, Promise) {
```
CookieMonsta89: Migrations look great.

index.js
```diff
@@ -0,0 +1,113 @@
...
+
+const sessionConfig = {
+    secret: 'who.throws.a.shoe.!',
+    name: 'monkey',
+    httpOnly: true,
+    resave: false,
+    saveUninitialized: false,
+    cookie: {
+        secure: false,
+        maxAge: 1000 * 60 * 1
```
CookieMonsta89: I believe normal practice would be to use this ```maxAge: 1 * 24 * 60 * 60 * 1000``` but it is really up to you

index.js
```diff
@@ -0,0 +1,113 @@
...
+    store: new KnexSessionStore({
+        tablename: 'sessions',
+        sidfieldname: 'sid',
+        knex: db,
+        createtable: true,
+        clearInterval: 1000 * 60 * 60
+    })
+};
+
+function protected(req, res, next) {
```
CookieMonsta89: Good work with you protected function.   Your endpoints look great as well.

PR #417, opened 2018-11-26
Kat Fries: first commit almost done

cpoIT:


api/server.js
```diff
@@ -0,0 +1,73 @@
...
+      ? res.status(200).json({ message: 'Logged in', user: user.username})
+      : res.status(401).json({ message: 'You shall not pass!' })
+      ;
+    })
+})
+
+
+server.get('/api/users', (req, res) => {
+  db('users')
+    .select('id', 'username', 'password')
```
kamry-bowman: obviously, you would never send the password back in the real world

PR #418, opened 2018-11-26
Sean Won 

swon8:


data/migrations/20181126115922_users.js
```diff
@@ -0,0 +1,16 @@
+
+exports.up = function(knex, Promise) {
```
CookieMonsta89: Migration looks correct and uniform.

index.js
```diff
@@ -0,0 +1,48 @@
+const express = require('express');
+const bcrypt = require('bcryptjs');
+const db = require('./data/db.js');
+const helmet = require('helmet');
+
```
CookieMonsta89: I need to see where you have added a protected function for sessions.

index.js
```diff
@@ -0,0 +1,48 @@
...
+server.use(helmet());
+
+server.get('/', (req, res) => {
+  res.send({ message: 'I am alive' });
+});
+
+//Register 
+server.post('/api/register', (req, res) => {
+  const creds = req.body;
+  const hash = bcrypt.hashSync(creds.password, 14);
```
CookieMonsta89: Good job hashing the passwords.

index.js
```diff
@@ -0,0 +1,48 @@
...
+  creds.password = hash;
+  db('users').insert(creds).then(id => res.status(201).json(id)).catch(err => res.json(err))
+})
+
+//Login
+server.post('/api/login', (req, res) => {
+  const creds = req.body;
+  db('users').where({ username: creds.username }).first().then(user => {
```
CookieMonsta89: I recommend making your endpoints for readable.  As they are, they seem too crammed together.  When using .where or .first or anything as such, try dropping them down a line to help with spacing.

PR #420, opened 2018-11-26
Ryan Walker - auth-i

rytwalker:
@kamry-bowman


index.js
```diff
@@ -0,0 +1,76 @@
...
+  const registrationData = req.body;
+  // validate data
+  if (!registrationData.username || !registrationData.password) {
+    return res
+      .status(400)
+      .json({ message: 'Please enter a valid username and password.' });
+  }
+  // Check if user already exists
+  try {
+    const userInDb = db.getByUsername(registrationData.username);
```
kamry-bowman: you need to `await` this if you want it to stop the rest of your code from running if the username already exists

index.js
```diff
@@ -0,0 +1,76 @@
...
+    return userInDb
+      ? res.status(422).json({ message: 'That username already exists.' })
+      : null;
```
kamry-bowman: this isn't a great pattern, because it returns no matter what, but only sends a http reply for one fork of the conditional. You should ensure you're returning a HTTP response of one kind or another. I would do this instead:
```
if (userInDb) {
    return res.status(422).json(...etc)
}
```

kamry-bowman: it still probably works in your code because you don't await the promise

index.js
```diff
@@ -0,0 +1,129 @@
...
+    res.status(500).json({ error: 'Something went wrong, pal.' });
+  }
+});
+
+// LOGOUT
+server.get('/api/logout', (req, res) => {
+  if (req.session) {
+    req.session.destroy(err => {
+      return err
+        ? res.json({ message: 'Something went wrong', err })
```
kamry-bowman: I would send an error code here

PR #422, opened 2018-11-26
Chad Jemmett Auth-i

ceejaay:
@sumi419


index.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+}
+
+server.get('/api/users', protected, (req, res) => {
+      db('users')
+        .then(users => res.status(200).json(users))
+        .catch(err => res.status(500).json(err))
+})
+
+server.post('/api/users', (req, res) => {
```
Velsu: That route is a little misleading. Better aproach would be to call it `register`.

index.js
```diff
@@ -0,0 +1,91 @@
...
+}
+
+server.use(session(sessionConfig))
+server.use(express.json())
+
+server.get('/', (req, res) => {
+  res.send(`Hello users number: ${req.session.user}`)
+})
+
+protected = (req, res, next) => {
```
Velsu: Nice middleware implementation.

index.js
```diff
@@ -0,0 +1,91 @@
...
+const knexConfig = require('./knexfile.js')
+const db = knex(knexConfig.development)
+const bcrypt = require('bcryptjs')
+const server = express();
+
+const sessionConfig = {
+  name: 'monkey',
+  secret: 'hello world',
+  cookie: {
+    maxAge: 1000 * 60 * 20 * 2,
```
Velsu: Proper usage of maxAge. Usually students make mistakes here and make them too short.

PR #428, opened 2018-11-26
David Soto - auth i

David247365:


David247365:
@CookieMonsta89

index.js
```diff
@@ -0,0 +1,113 @@
...
+const bcrypt = require("bcryptjs");
+const session = require("express-session");
+const KnexSessionStore = require("connect-session-knex")(session);
+
+const db = require("./database/dbConfig.js");
+
+const server = express();
+
+// Remeber that this is a secret
+const sessionConfig = {
```
CookieMonsta89: Session config looks good.

index.js
```diff
@@ -0,0 +1,113 @@
...
+    .catch(err =>
+      res.status(500).json({ message: "There was an error, please try again" })
+    );
+});
+
+// Get users
+
+// Protect this route and only allow authenticated users to see it
+server.get("/api/users", (req, res) => {
+  if (req.session && req.session.userId) {
```
CookieMonsta89: You can create a function that does this for you.  ```function protected(req, res, next) {
    if (req.session && req.session.username) {
        next();
    } else {
        return res.status(401).json({error:'Incorrect Credentials'});
    }
}```

CookieMonsta89: Makes it reusable

PR #435, opened 2018-11-26
Hunter Smith - auth-i

Hunter315:
@kamry-bowman


index.js
```diff
@@ -0,0 +1,66 @@
...
+
+    //override the user.password with the hash
+        creds.password = hash;
+
+    //save the user to the database
+        db('users').insert(creds)
+        .then(ids => {
+        res.status(201).json(ids);
+        })
+        .catch(err => res.status(401).json(err));
```
kamry-bowman: 401 is not authorized, it's not clear if the failure was due to that reason. Could be that they didn't send in a password for example

index.js
```diff
@@ -39,28 +63,55 @@ server.post('/api/register', (req, res) => {
...
-});
- 
+      .catch(err => res.send(err));
+  });
+  
+  server.get('/api/logout', (req, res) => {
+    if(req.session) {
+      req.session.destroy(err => {
+        if (err) {
+          res.send('you can never leave');
```
kamry-bowman: you should send an error code

index.js
```diff
@@ -39,28 +63,55 @@ server.post('/api/register', (req, res) => {
...
+        } else {
+          res.send('bye');
```
kamry-bowman: you should send a success code

PR #436, opened 2018-11-26
Adam Hinckley

adamhinckley:
…dpooint not working


index.js
```diff
@@ -0,0 +1,107 @@
...
+});
+
+server.post("/api/login", (req, res) => {
+  // grab username and password from body
+  const creds = req.body;
+  db("users")
+    .where({ username: creds.username })
+    .first()
+    .then(user => {
+      if (user && bcrypt.compareSync(creds.password, user.password)) {
```
Velsu: It is good idea to delegate bcrypt to middleware files and create a separate method for it to use, but not necessary.

PR #442, opened 2018-11-26
Gear - Intro to Authentication

sabrinagear:


KevinBrack:
Looks like this is coming along well. A few notes however. On line 36 there is an empty call to `db("users");` and that endpoint is missing the definition of `creds`. I assume you want that to point to the `req.body` the way you are using it. That is blocking me from logging in.

Also I was able to get the users from the database without logging in. Need to secure that with some middle-ware or some form of validation.

PR #443, opened 2018-11-26
Yanrong_Xu - auth-i

YanrongXu:
@kamry-bowman


index.js
```diff
@@ -0,0 +1,80 @@
...
+            res.status(500).json(err);
+        })
+});
+
+server.post('/api/login', (req, res) => {
+    const creds = req.body;
+
+    db('users').where({username: creds.username}).first().then(user => {
+        if(user && bcrypt.compareSync(creds.password, user.password)) {
+            req.session.username = user.username
```
kamry-bowman: I think it would be more common to use the user Id

index.js
```diff
@@ -77,4 +90,18 @@ server.post('/api/login', (req, res) => {
     })
 });
 
+server.get('/api/logout', (req, res) => {
+    if (req.session) {
+        req.session.destroy(err => {
+            if(err) {
+                res.send('LoL, still not logout yet')
```
kamry-bowman: I would still send back http status codes for this one and below

PR #454, opened 2018-11-26
auth-i Brandon Grych 

brandongrych:


index.js
```diff
@@ -0,0 +1,107 @@
...
+const server = express();
+
+const sessionConfig = {
+    secret: 'Extremelysecretthinghere',
+    name: 'brad',
+    httpOnly: true,
+    saveUninitialized: false,
+    cookie: {
+        secure: false,
+        maxAge: 1000 * 60 * 10
```
Velsu: It seems setting 10 minutes for cookie max age can sometimes cause issues as i was unable to login. Needed to alter the max-age to at least 1 hour. Just wanted to give headsup that it can be an issue if You are not able to view protected route and are already logged in.

index.js
```diff
@@ -0,0 +1,107 @@
...
+    .catch(err => {
+        res.status(500).json(err);
+    });
+});
+
+server.post('/api/login', (req,res) => {
+    const creds = req.body;
+    db('users').where({username: creds.username}).first()
+    .then(user => {
+        if(user && bcrypt.compareSync(creds.password, user.password)) {
```
Velsu: Sometimes it is a good idea to delegate bcrypt methods to a separate middleware file, create a function and call it in index after importing.

PR #458, opened 2018-11-26
Auth-I Edd Burke

bummings:


KevinBrack:
The requested endpoints in the instructions were
POST `/api/register`
POST  `/api/login`
GET  `/api/users`

Not a huge deal, but if you were working in a team and someone was relying on those endpoints for their code, it is not a request that would be open to change.

I was able to get the list of users with out logging in. Need to look into that.

On line 55 your `.catch` is inside the `.then` and threw an error when I registered because of that.

KevinBrack:
Great work. All requested changes were made. Server meets requirements, and hides the users behind authentication with sessions. Thank you.

PR #464, opened 2018-11-26
auth-i-ian-cameron

codypeak:


KevinBrack:
This is really great work Ian. Authentication, registering, and logging in worked flawlessly. `GET /api/users` however just returned to me `"users"`.

Did a little digging and I found the problem in `index.js` on line 65
`res.json('users');` is in quotes so it just sends that string instead  of the contents of the variable. Should be `res.json(users);`.

After that small change it works flawlessly. Great job sir.

PR #468, opened 2018-11-27
Auth I [Chance Embrey]

cembreyfarquhar:


index.js
```diff
@@ -0,0 +1,79 @@
...
+const KnexSessionStore = require("connect-session-knex")(session);
+const bcrypt = require("bcryptjs"); // *************************** added package and required it here
+
+const db = require("./database/dbConfig.js");
+
+const server = express();
+
+const sessionConfig = {
+  name: "monkey",
+  secret: "asfjaofuwruq04r3oj;ljg049fjq30j4jlajg40j40tjojasl;kjg",
```
Velsu: You want to try and store secrets in env variable so it is not accessible for everyone.

index.js
```diff
@@ -0,0 +1,79 @@
...
+
+server.post("/api/register", (req, res) => {
+  const creds = req.body;
+
+  const hash = bcrypt.hashSync(creds.password, 4);
+
+  creds.password = hash;
+
+  db("users")
+    .insert(creds)
```
Velsu: You lack any kind of check if password or username was passed within body. Also no proper error handling.

index.js
```diff
@@ -0,0 +1,79 @@
...
+        res.status(200).json({ message: "Welcome! " });
+      } else {
+        res.status(401).json({ message: "can't login" });
+      }
+    })
+    .catch(err => res.json(err));
+});
+
+server.get("/api/users", (req, res) => {
+  if (req.session && req.session.user) {
```
Velsu: There is no fallback in the case user is not logged in, server will just go into infinite loading and end user won't receive any response.

PR #475, opened 2018-12-03
Angelo Deleon - auth-i

RaDeleon:
William PM - @Wdifulvio523


index.js
```diff
@@ -0,0 +1,129 @@
...
+
+    // generate a hash
+    const hash = bcrypt.hashSync(registrationData.password, 10);
+
+    // replace plain text with hash
+    registrationData.password = hash;
+
+    // save to db
+    try {
+        const newUserId = await db.addUser(registrationData);
```
Velsu: You are trying to call a function `addUser` which does not exist, because of that it is impossible to add new user.

PR #479, opened 2018-12-03
auth -i / Govi Escudero

jaklundun:


index.js
```diff
@@ -0,0 +1,149 @@
...
+ function protected(req, res, next) {
+    if (req.session && req.session.user) {
+      next();
+    } else {
+      res.status(401).json({ you: 'Access Denied!' });
+    }
+  }
+
+  // express-session middleware
+  function protected(req, res, next) {
```
Velsu: Duplicate of protected method.

index.js
```diff
@@ -0,0 +1,149 @@
...
+      sidfieldname: 'sid',
+      knex: db,
+      createtable: true,
+      clearInterval: 1000 * 60 * 60,
+    }),
+  };
+
+  server.use(session(sessionConfig)); // hooks up express-session to our api
+
+const server = express();
```
Velsu: You need to define server at the top of the file.

index.js
```diff
@@ -0,0 +1,149 @@
...
+
+//Initialize db
+const db = knex(knexConfig.development);
```
Velsu: Same as for server, that needs to go on top of the file or it will be undefined.

index.js
```diff
@@ -0,0 +1,149 @@
...
+    
+    const creds = req.body;
+  
+    db('users')
+      .where({ username: creds.username })
+      .first()
+      .then(user => {
+        if (user && bcrypt.compareSync(creds.password, user.password)) {
+         
+          res.status(200).json({ message: 'Benvenuto!' });
```
Velsu: You are not setting a session after login, hence the protected method will not let anyone view the contents even after logging in.

PR #482, opened 2018-12-04
barbara hunter

cmdncode:


index.js
```diff
@@ -0,0 +1,123 @@
...
+
+const server = express();
+
+
+//session cookie
+const sessionConfig = {
+  name: 'the cookie monster',
+  secret: process.env.SECRET,
+  cookie: {
+    maxAge: 1000 * 60 * 20,
```
Velsu: There is a major bug in here that set's too short lifespan of a cookie. With current age cookie will be outdated instantly and req.session.userId will be undefined.

Change that to `maxAge: 1000 * 60 * 20 * 12`

index.js
```diff
@@ -0,0 +1,123 @@
...
+}
+
+server.use(session(sessionConfig));
+server.use(express.json());
+server.use(helmet());
+server.use(cors());
+
+
+//middleware for restricting specific routes
+const checkUser = (req, res, next) => {
```
Velsu: Nice middleware.

PR #485, opened 2019-01-16
Christopher Ferrell - auth-i

cferrell2:
auth-i  Project
@juliankohlman

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 ## Assignment
 
-**Part one, due Monday**: Use Node.js, Express and Knex to build an API that provides **Register** and **Login** functionality using SQLite to store _User_ information. Make sure the password is not stored as plain text.
+**Part one, due Monday**: Use Node.js, Express and Knex to build an API that provides **Register** and **Login** functionality using SQLite to store _User_ information. Make sure the password is not stored as plain text.-
 
 **Part two, due Tuesday**: Use **sessions** and **cookies** to keep a record of logged in users across requests.
 


Routers/usersRouter.js
```diff
@@ -0,0 +1,77 @@
+const userDb = require('../data/helpers');
+
```
juliankohlman: This looks great Christopher 👍 code is clean, and comments are acting as documentation.

Routers/usersRouter.js
```diff
@@ -0,0 +1,77 @@
...
+//create router
+const express = require('express');
+router = express.Router();
+
+//grab encryption library 
+const bcrypt = require('bcryptjs');
+
+//grab checkSession middleware function
+const verify = require('../middleware');
```
juliankohlman: Could experiment with destructuring your middleware once the file begins to grow 👍 
```suggestion
const { checkSession } = require('../middleware');
```

Routers/usersRouter.js
```diff
@@ -0,0 +1,77 @@
...
+    .catch(err =>{
+        res.status(500).json({error:"Unable to retrieve users"})
+    })
+})
+
+//Log out user
+router.get('/logout', (req,res) =>{
+    console.log(req.session);
+    if(req.session){
+        req.session.destroy(err =>{
```
juliankohlman: Good find with `.destroy(callback)` from the documentation 👍 Always nice to leverage libraries where we can.

PR #487, opened 2019-01-16
Bryce Evans

BryceEvans:


database/dbHelpers.js
```diff
@@ -0,0 +1,17 @@
+const knex = require('knex');
+
+const knexConfig = require('../knexfile');
+
+const db = knex(knexConfig.development);
+
+// this file contains my helpers which is basically the knex logic.
```
kiadorii: I would suggest putting this comment on the top of the file for formatting.

PR #490, opened 2019-01-16
Julie-Jonak: auth-i

juliejonak:
@katiagilligan888 Opening pull request to start project


data/seeds/01_users.js
```diff
@@ -0,0 +1,12 @@
+
+exports.seed = function(knex, Promise) {
+  return knex('Users').truncate()
+    .then(function () {
+      return knex('Users').insert([
+        {Username: "UserOne", Password: "1234567zyxw"},
```
katiagilligan888: Do not suggest seeding passwords or in this case user accounts, because this information is not secure without hashing

index.js
```diff
@@ -0,0 +1,102 @@
...
+server.get('/', (req, res) => {
+    res.json({ message: "Welcome"})
+})
+
+//Test outside of routes to let sessions work
+
+
+server.post('/api/register', (req, res) => {
+    const user = req.body;
+    if(user.Password.length > 12){
```
katiagilligan888: For an added bonus, look into a validation library for node

index.js
```diff
@@ -0,0 +1,102 @@
...
+            dbLogin.add(user)
+            .then(response => {
+                res.status(201).json({ message: "Account created successfully!" })
+            })
+            .catch(err => {
+                res.status(500).json({ message: "Unable to add new account" })
+            })
+        } else{
+            res.status(400).json({ message: "New accounts require a Username" })
+        }
```
katiagilligan888: What about checking to see if the username already exists? Will the db overwrite an existing user?

index.js
```diff
@@ -0,0 +1,102 @@
...
+        res.status(400).json({ message: "Password must be at least 12 characters long"})
+    }
+})
+
+server.post('/api/login', (req, res) => {
+    const loginUser = req.body;
+    if(loginUser.Username && loginUser.Password){
+        dbLogin.login(loginUser.Username)
+            .then(response => {
+                console.log('session', req.session, 'id', response[0].id)
```
katiagilligan888: Remember to always comment out console log statements before committing any code. Another way to take care of console.log statements is in a function to have a debug variable that is set to true. Then wrap your console.log statements in an `if(debug){console.log("Hello")}` and then when finished testing set debug back to false

knexfile.js
```diff
@@ -0,0 +1,51 @@
+// Update with your config settings.
+
+module.exports = {
```
katiagilligan888: What is the use case for the staging and production in the knexfile? What is a good example for both?

PR #491, opened 2019-01-16
Jesse Anderson - auth-i

Jesse9009:
@juliankohlman


dbHelpers.js
```diff
@@ -0,0 +1,20 @@
+const express = require('express');
+const db = require('./dbConfig');
+
+module.exports = {
```
juliankohlman: Placing your exports on the top here will work because of function declaration hoisting. I've mostly seen it placed on the bottom of files, but at the end of the day it will depend on your team's preferences, but as long as you're consistent with placement, and you know why this works you're good 👍  
You could also bundle up the whole thing if the file starts to get long
```javascript
module.exports = {  
    helperName: function() {}
    helperName: function() {}
}
```

index.js
```diff
@@ -0,0 +1,101 @@
...
+const cors = require('cors');
+const bcrypt = require('bcryptjs');
+const session = require('express-session');
+
+const db = require('./dbHelpers');
+const restrictedRoutes = require('./restrictedRoutes');
+
+const server = express();
+
+// Custom Middleware
```
juliankohlman: Nice and simple labeling, but super helpful 👍

index.js
```diff
@@ -0,0 +1,101 @@
...
+    resave: false,
+    saveUninitialized: false
+  })
+);
+
+server.get('/', (req, res) => {
+  res.send('Live server!');
+});
+
+server.use('/api/restricted', restrict, restrictedRoutes);
```
juliankohlman: Good job building out your server with middleware Jesse 🔥 this line sort of encapsulates how middleware adds more features/abilities to our APIs.

PR #494, opened 2019-01-16
Denise Escobar Auth-i

DenEscobar:
@katiagilligan888


data/migrations/20190116125500_users_table.js
```diff
@@ -0,0 +1,16 @@
+
+exports.up = function(knex, Promise) {
+    return knex.schema.createTable('users', users =>{
+        users.increments();
+
+        users
+            .string('username', 128)
+            .notNullable()
+            .unique();
```
katiagilligan888: Good job adding unique as a users property to verify that no username is used twice 👍

index.js
```diff
@@ -0,0 +1,98 @@
...
+        .json({errorMessage: "Please provide your username and password"})
+    }
+
+}); 
+
+//GET	/api/users
+
+server.get('/api/users', (req, res) =>{
+    console.log(req.session)
+    if(req.session && req.session.id){
```
katiagilligan888: Instead of the if statement on line 81 you could create middleware that can be useful for more than one protected endpoint under a signed in user with sessions

knexfile.js
```diff
@@ -0,0 +1,48 @@
+// Update with your config settings.
+
+module.exports = {
+
```
katiagilligan888: Do you know the use cases for the staging and production in the knexfile?

PR #496, opened 2019-01-16
Lee Lichtsinn / auth-i

leelichtsinn:
installed dependencies
created basic server
@47analogy


leelichtsinn:
added users schema
added dbConfig
added register route to create new user
added login route to check password

leelichtsinn:
added new user cookie to login endpoint

leelichtsinn:
added session configuration
refactored login endpoint
added GET endpoint for users
added custom middleware to check user's credentials
passed protect middleware as an arg in users endpoint

PR #498, opened 2019-01-16
Michael Littleton - Auth I

mglittleton:


data/migrations/20190116214902_users.js
```diff
@@ -0,0 +1,14 @@
+exports.up = function(knex, Promise) {
+  knex.schema.createTable("users", table => {
+    table.increments();
+    table
```
Raymondgrc: Good styling and schema set up.

index.js
```diff
@@ -0,0 +1,36 @@
+const express = require("express");
```
Raymondgrc: I also like the division between imports and server settings here.

index.js
```diff
@@ -0,0 +1,36 @@
...
+const PORT = 4200;
+
+const server = express();
+
+server.use(express.json(), cors());
+
+// | POST   | /api/register | Creates a `user` using the information sent inside the `body` of the request. Hashes the password
+
+server.post("./api/register", (req, res) => {
+  const user = req.body;
```
Raymondgrc: I would recommend destructuring this object into an appropriate variable but this is just preference.

index.js
```diff
@@ -0,0 +1,36 @@
...
+  user.password = bcrypt.hashSync(user.password, 12);
```
Raymondgrc: It may be tedious, but I recommend adding a server side check to be sure that the user provided a username and a password.

PR #503, opened 2019-01-16
 Chris Roth Auth-i:

TexaCaliCoder:
initialized yarn and knex
added dependancies: express, knex, sqlite3, morgan, and nodemon(dev)
built out boilerplate for express knex api in index.js
added users table with knex migrations
- @katiagilligan888 initial commit for PR


data/helpers/validations.js
```diff
@@ -0,0 +1,7 @@
+const Joi = require('joi')
+module.exports={
```
katiagilligan888: Good job using JOI! Glad you like it!

routes/loginRoutes.js
```diff
@@ -0,0 +1,48 @@
...
+                res.status(406).json({err: "Invalid Username or Password"});
+            }
+        })
+        .catch(err=>{
+            res.status(500).json({error:err})
+        })
+})
+
+router.post('')
+
```
katiagilligan888: Add a get request that is a protected route that only signed in users can see. This protected endpoint can be built out with a middleware function.

routes/loginRoutes.js
```diff
@@ -0,0 +1,48 @@
+const express = require('express');
+const db = require('../data/helpers/helpers');
+const Joi = require('joi');
+const validate = require('../data/helpers/validations');
+
+const router = express.Router();
+const bcrypt = require('bcryptjs');
+
```
katiagilligan888: Make sure to incorporate sessions in the server as that is the main focus of the assignment 👍

PR #504, opened 2019-01-16
Levi Appenfelder

LeviApp:
@katiagilligan888


data/seeds/01_users.js
```diff
@@ -0,0 +1,13 @@
+
+exports.seed = function(knex, Promise) {
+  // Deletes ALL existing entries
+  return knex('users').truncate()
+    .then(function () {
+      // Inserts seed entries
+      return knex('users').insert([
+        {username: 'Crash', password: 'UkaUka', registered: true},
+        {username: 'Groot', password: 'IamGroot', registered: true},
```
katiagilligan888: I suggest that you do not seed user accounts into the database because security is compromised since there is no hashing occurring.

server.js
```diff
@@ -0,0 +1,81 @@
...
+    }, 
+    httpOnly: true, // don't let JS code access cookies. Browser extensions run JS code on your browser!
+    resave: false,
+    saveUninitialized: false,
+  }))
+
+const PORT = 4444;
+
+server.get('/api/users', (req, res) => {
+    if (req.session && req.session.userID){
```
katiagilligan888: Instead of the if statement on line 22, you can create a middleware function that is only allows users with sessions signed in to access a get endpoint. That way if you have multiple endpoints that are restricted to signed in users, you can re use the middleware

server.js
```diff
@@ -0,0 +1,81 @@
...
+        
+        const user = req.body;
+        
+        DB.findByUsername(user.username)
+        .then(users => {
+            if (users && bcrypt.compareSync(user.password, users[0].password)){
+                req.session.userID = users[0].id;
+                res.json({message: `Success!`})}
+            else {
+                res.status(404).json({message: `invalid username or password`})
```
katiagilligan888: Small picky critique: Try to keep your JSON message responses consistent in terms of capitalization for professionalism.

PR #507, opened 2019-01-16
Rob Salzberg - auth-i

robsalzberg:
@juliankohlman

- [x] Set-up project
- [x] Create POST register endpoint
- [x] Create POST login endpoint
- [x] Create GET users endpoint
- [x]  Create sessions and cookies


dbHelpers.js
```diff
@@ -0,0 +1,24 @@
...
+function insertUser(user) {
+  return db("users").insert(user);
+}
+
+function findByUsername(username) {
+  return db("users").where("username", username);
+}
+function findUsers() {
+  return db("users").then(response => {
+    const mapped = response.map(x => {
```
juliankohlman: I'd use a more descriptive variable inside your map or a good old `item`, or `element` to keep it vanilla 👍

index.js
```diff
@@ -0,0 +1,57 @@
...
+  } else {
+    res.status(400).json({ message: "Please enter a username and password" });
+  }
+});
+
+server.post("/api/login", (req, res) => {
+  const user = req.body;
+  db.findByUsername(user.username)
+    .then(dbUser => {
+      if (dbUser[0] && bcrypt.compareSync(user.password, dbUser[0].password)) {
```
juliankohlman: This credential check looks good Rob 👍 you're checking for the existence of the user and for a matching password hash.

PR #508, opened 2019-01-16
Jonathan Brock

jbrock1981:
@FireInjun initial PR


jbrock1981:
## Assignment

- [ ] **Part one, due Monday**: 

    Use Node.js, Express and Knex to build an API that provides **Register** and **Login** functionality 
    using SQLite to store _User_ information. Make sure the password is not stored as plain text.

- [ ] **Part two, due Tuesday**: 

    Use **sessions** and **cookies** to keep a record of logged in users across requests.

### Implement Requirements

- [ ] - Take the steps necessary to create a `package.json` to keep a record of all dependencies.
- [ ] - Configure an _npm script_ named _"start"_ that will execute your code using _nodemon_ so that the server restarts on changes. Make _nodemon_ be a development time dependency only, it shouldn't be deployed to production.

- [ ] - Design and build a set of endpoints listed below.
- [ ] - **Use _Postman_ to test the API as you work through the exercises.**

#### Endpoints

- [ ] | POST   | /api/register | Creates a `user` using the information sent inside the `body` of the request. 
**Hash the password** before saving the user to the database.                                                                                                                                                 

- [ ] | POST   | /api/login    | Use the credentials sent inside the `body` to authenticate the user. On successful login, create a new session for the user and send back a 'Logged in' message and a cookie that contains the user id. If login fails, respond with the correct status code and the message: 'You shall not pass!' |

- [ ] | GET    | /api/users    | If the user is logged in, respond with an array of all the users contained in the 
database. If the user is not logged in respond with the correct status code and the message: 'You shall not pass!'.            |

## Stretch Problem

- [ ] - Write a piece of **global** middleware that ensures a user is logged in when accessing _any_ route 
prefixed by `/api/restricted/`. For instance, `/api/restricted/something`, `/api/restricted/other`, and `/api/restricted/a` should all be protected by the middleware; only logged in users should be able to access these routes.

- [ ] - Build a React application that implements components to register, login and view a list of users. Gotta keep sharpening your React skills.

index.js
```diff
@@ -0,0 +1,88 @@
...
+const session = require('express-session');
+const server = express();
+
+const PORT = 4200;
+
+server.use(express.json());
+
+server.use(session({
+    name: 'sessionTest',
+    secret: 'SECRET',
```
katiagilligan888: I would not suggest using such an obvious secret. It should be something that should not be easily hacked

index.js
```diff
@@ -0,0 +1,88 @@
...
+}));
+
+function protect(req, res, next){
+    if (req.path.includes('/restricted/') || req.path == '/api/users') {
+        if (req.session && req.session.userId) {
+            next();
+        } else {
+            res.status(401).json({ message: 'go away' })
+        }
+        } else {
```
katiagilligan888: On line 29 adjust indentation

index.js
```diff
@@ -0,0 +1,88 @@
...
+    .select('id', 'username')
+    .then(users => {
+        res.json(users);
+    })
+    .catch(err => res.json(err));
+});
+
+server.post('/api/register', (req, res) => {
+    const creds = req.body;
+    db('users')
```
katiagilligan888: Add validation here for the creds. You should validate that the username and password are both present and meet certain criteria before making the call to the database.

index.js
```diff
@@ -0,0 +1,88 @@
...
+        .where({ username: creds.username })
+        .first()
+        .then(user => {
+            if (user) {
```
katiagilligan888: This if statement on line 55 may not be necessary because in your migrations you already specified that a username has to be unique, which means the db will not let you create another user automatically with the same username

index.js
```diff
@@ -0,0 +1,88 @@
...
+            }
+        })
+        .catch(err => res.json(err));
+})
+
+server.post('/api/login', (req, res) => {
+    const creds = req.body;
+    db('users')
+        .where({ username: creds.username })
+        .first()
```
katiagilligan888: No need to user first() here if there is only 1 user with username

PR #509, opened 2019-01-16
Dewayne Lindsay auth-i @raymondgrc

Dewayne87:
finished adding endpoints


data/migrations/20190116144803_users_table.js
```diff
@@ -0,0 +1,18 @@
+
+exports.up = function (knex) {
+    return knex.schema.createTable('users', table => {
+        table
```
Raymondgrc: Good job setting up a schema here. I would recommend a different styling here because this can get tedious to read if there were more tables.

package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "auth-i",
+  "version": "1.0.0",
+  "main": "index.js",
+  "scripts": {
+    "start": "nodemon index.js"
```
Raymondgrc: index.js is not required, in case you did not know. There is no harm in keeping it there though :)

index.js
```diff
@@ -0,0 +1,62 @@
...
+        })
+        .catch(error => {
+            res
+                .status(500)
+                .json(error)
+        })
+})
+
+server.listen(PORT, err => {
+    if (err) console.log(err);
```
Raymondgrc: I like that you're checking for errors here. I don't see it often. It's a good practice.

dbhelper.js
```diff
@@ -0,0 +1,20 @@
+const knex = require('knex');
+const dbconfig = require('./knexfile');
+const db = knex(dbconfig.development);
+
+registerUser = (newUser) => {
+    return db('users')
+        .insert(newUser).
```
Raymondgrc: Your helpers are nice. Just try to watch for coding inconsistencies like this. They can be a trouble fixing later when working with a team. the . method should be completely on a new line.

index.js
```diff
@@ -0,0 +1,62 @@
+const express = require('express');
```
Raymondgrc: I also recommend splitting these one liners into more organized sections. Like a section for imports, a section for implementing imports, a section for settings, separated by a new line.

index.js
```diff
@@ -0,0 +1,62 @@
...
+const server = express();
+const logger = require('morgan');
+const helmet = require('helmet');
+const PORT = 4000;
+const dbhelper = require('./dbhelper')
+const bcrypt = require('bcryptjs');
+server.use(express.json());
+server.use(helmet());
+server.use(logger('dev'))
```
Raymondgrc: Stick to either semicolons or no semicolons.

index.js
```diff
@@ -0,0 +1,62 @@
...
+server.disable("etag");
+
+server.post('/api/register', (req, res) => {
+    const newUser = req.body;
```
Raymondgrc: Instead of accessing this objects properties later, it might be helpful to destructure this object into variables that you'll use later.

index.js
```diff
@@ -0,0 +1,62 @@
...
+    console.log(newUser)
+    if (!newUser.password || !newUser.username) {
+        return res
+            .status(404)
```
Raymondgrc: There are more appropriate status codes that could be provided. I would suggest checking this https://www.restapitutorial.com/httpstatuscodes.html out.

index.js
```diff
@@ -0,0 +1,62 @@
...
+            .json('Please provide Username and/or password')
+    }
+    newUser.password = bcrypt.hashSync(newUser.password, 14);
```
Raymondgrc: I would make this into another variable. Overwriting the variable stops it from being used and it might be helpful to have this variable remain pure during this process.

PR #510, opened 2019-01-17
Angelina La Salle auth-i

alasalle:
@juliankohlman


customMiddleware.js
```diff
@@ -0,0 +1,20 @@
...
+const auth = (req, res,next) => {
+  if (req.session && req.session.userId) {
+    next();
+  } else {
+    res.status(400).send("Restricted!");
+  }
+}
+
+module.exports = {
+  protect: protect,
```
juliankohlman: You can use object-shorthand here 👍 
```suggestion
module.exports = { protect, auth }
```

routers/usersRouter.js
```diff
@@ -0,0 +1,95 @@
+const usersDb = require("../data/helpers/usersDb");
+const bcrypt = require("bcryptjs");
+const customMw = require("../customMiddleware");
+
+const express = require("express");
+const router = express.Router();
+const protect = customMw.protect;
+const auth = customMw.auth;
+
```
juliankohlman: You could grab these from you `customMw` with destructuring 
```javascript 
const { protect, auth} = customMw
```
You're still passing in the `protect` over `customMw.protect`to your endpoints so the above suggestion would be more useful as you have more middleware to deal with 👍

routers/usersRouter.js
```diff
@@ -0,0 +1,95 @@
...
+// router.use("/restricted", auth())
+
```
juliankohlman: Really like that, you're checking for correct data types, and the absence of data as well 🔥 Check out this [article](https://medium.freecodecamp.org/how-to-make-input-validation-simple-and-clean-in-your-express-js-app-ea9b5ff5a8a7) about input validation in express. There many ways to get this done, but hopefully this will get you thinking about how you can do this with middleware 👍

PR #513, opened 2019-01-17
Robert Billington, FSWPT2, PM @katiagilligan888; Initial Commit. Set …

chillingtonbillington:
…up package.json and installed dependencies


data/seeds/users.js
```diff
@@ -0,0 +1,11 @@
+
+exports.seed = function(knex, Promise) {
+  // Deletes ALL existing entries
+  return knex('users').truncate()
+    .then(function () {
+      // Inserts seed entries
+      return knex('users').insert([
+        {username: 'test-name', password: 'test-password'}
```
katiagilligan888: I would suggest to not add seeding for your user account data because the passwords are not hashed which compromises security.

PR #520, opened 2019-01-17
Michael Trew Auth-I

mtrew2015:
@juliankohlman


database/dbHelpers.js
```diff
@@ -0,0 +1,15 @@
+const knex = require('knex');
+const knexConfig = require ('../knexfile.js');
+const db = knex(knexConfig.development)
+
```
juliankohlman: I'm glad you caught that typo here Michael 👍 Your config and helpers look good.

index.js
```diff
@@ -0,0 +1,75 @@
+const express = require('express');
+const server = express()
+const cors = require('cors')
+const db = require('./database/dbHelpers.js')
+const bcrypt = require('bcryptjs');
+const session = require('express-session');
+const middleWare = require('./middleware.js')
```
juliankohlman: Nice job bringing in your protected route middleware. If you end up extending you middleware you could shift to destructuring that object 
```javascript
const { protected, middleware1, middleware2, etc } = require('./middleware.js')
```
then when you pass your middleware to the endpoints it will look like 
```javascript
server.get('/api/users', protect, (req, res) => {})
```

index.js
```diff
@@ -0,0 +1,75 @@
...
+    .then(id => {
+      res.status(201).json({ id: id[0] })
+    })
+    .catch(err => {
+      res.status(500).json({ error: 'Please Try Again' })
+    });
+});
+
+server.post('/api/login', (req, res) => {
+  const creds = req.body;
```
juliankohlman: I'd consider tweaking this variable name to read as `credentials` or `user` for clarity 👍

PR #521, opened 2019-01-17
Luis Alvarez

TheGameFreak720:
@juliankohlman


routes/users.js
```diff
@@ -0,0 +1,51 @@
...
+            res.status(500).json({ errorMessage: 'Failed to insert user' });
+        }) 
+});
+
+router.post('/login', (req, res) => {
+   const user = req.body;
+   
+   userDB.get(user)
+    .then(users => {
+        if(users.length && bcrypt.compareSync(user.password, users[0].password)) {
```
juliankohlman: Like the usage of boolean `&&` operator here to check the users' array and the password within a single if statement 🔥

PR #535, opened 2019-01-20
Jorge Osto - Finished all endpoints

Jorlost19:
added session persistance after server restarts
added middlewaare to access protected routes


index.js
```diff
@@ -0,0 +1,82 @@
...
+
+server.post('/api/register', (req, res) => {
+  const user = req.body;
+  if (!user.password || !user.username) {
+    res.status(400).json({ error: 'Username and Password are required!' });
+  }
+  user.password = bcrypt.hashSync(user.password, 12);
+  db('users')
+    .insert(user)
+    .then(ids => res.status(201).json(ids[0]))
```
johnoro: You could [destructure](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Array_destructuring) the first element in your `then` callback:
```javascript
.then(([ id ]) => res.status(201).json(id))
```

index.js
```diff
@@ -0,0 +1,82 @@
...
+});
+
+server.post('/api/login', (req, res) => {
+  const creds = req.body;
+  db('users')
+    .where('username', creds.username)
+    .first()
+    .then(user =>
+      user && bcrypt.compareSync(creds.password, user.password)
+        ? ((req.session.userId = user.id), res.json({ message: 'Logged in!' }))
```
johnoro: Why not have some sort of a success status here as well? Typically you'll see something like [302](https://en.wikipedia.org/wiki/HTTP_302) or at least just a 200 (for OK) here.

Jorlost19:
Hello John, thanks for looking at my code, in the login endpoint I didnt
send back any status because status code 200 is the default status, so when
you don't send any status, it will send the 200 for you, so thats why I
didnt, because it wasn't necessary. Again thanks for your time.

PR #544, opened 2019-01-24
Humberto Raya auth-i

hraya:
@johnoro


index.js
```diff
@@ -0,0 +1,57 @@
...
+
+
+server.post('/api/register', (req, res) =>{
+    const user = req.body;
+    const hash = bcrypt.hashSync(user.password, 14);
+    user.password = hash;
+    if(!user.username || !user.password){res.status(400).json({message:'Please provide username and password!'})}
+    db.insertUser(user)
+    .then(ids =>{
+        const id = ids[0]
```
johnoro: There are a couple of alternative ways to get that `id` off of the array:
- [Array destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Array_destructuring):
```javascript
db.insertUser(user)
  .then(([ id ]) => {
    res.json({ newUserId: id })
  })
  // ...
```
- the [first](https://knexjs.org/#Builder-first) method
```javascript
db.insertUser(user)
  .first()
  .then(id => {
    res.json({ newUserId: id })
  })
  // ...
```

index.js
```diff
@@ -0,0 +1,57 @@
...
+        if(user && bcrypt.compareSync(creds.password, user.password)){
+            res.json({welcome:user.username})
+        }else{
+            res.status(401).json({message:"Invalid username or password!"})
+        }
+    })
+    .catch(err =>{
+        res.status(err =>{
+            res.status(500).json({err})
+        })
```
johnoro: This code block looks a little odd; why are you trying to pass a callback function to the `status` method? I'm assuming you wanted lines 41-45 to look a bit more like this:
```javascript
.catch(err => {
  res.status(500).json({ err });
});
```

index.js
```diff
@@ -0,0 +1,57 @@
...
+    })
+})
+
+server.get("/api/users", (req, res) => {
+    db.getUsers()
+      .then(users => {
+        res.json(users);
+      })
+      .catch(err => res.send(err));
```
johnoro: You should explicitly send a status. I believe it defaults to 200 ("OK"), which certainly isn't what you want for an error:
```javascript
db.getUsers()
  .then(users => {
    res.json(users); // it'd be a good idea to be explicit here as well
  })
  .catch(err => res.status(500).json({ err }));
```

index.js
```diff
@@ -1,14 +1,36 @@
...
 
+const sessionConfig = {
+    secret:'This-is-a-secret!',
+    name:'superMonkey',
+    httpOnly:true,
+    resave:false,
+    saveUninitialized:false,
+    cookie:{
+        secure:false,
+        maxAge: 1000 * 60 *01
```
johnoro: Is the leading `0` on the `01` here on purpose?

index.js
```diff
@@ -1,14 +1,36 @@
 const express = require('express');
 const cors = require('cors');
-
 const bcrypt = require('bcryptjs');
+const session = require('express-session');
+const protects = require('./middleware.js')
```
johnoro: Since you only have once middleware function in this 'protects' object, you could just [destructure](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring) it off:
```javascript
const { protected } = require('./middleware');
// ...
```
and then, later on down in your code:
```javascript
server.get("/api/users", protected,  (req, res) => {
  // ...
```

PR #548, opened 2019-01-30
[Tommy Collison] Auth-i

tommycollison:
@johnoro 🔏 🔐


tommycollison:
MVP complete!
- [x] POST: /api/register
- [x] POST: /api/login
- [x] GET | /api/users

index.js
```diff
@@ -0,0 +1,54 @@
...
+        } else {
+            res.status(404).json(`You shall not pass!`)
+        }
+    })
+    .catch(err => {res.status(500).send(err)})
+})
+
+
+server.post('/api/login', (req, res) => {
+})
```
johnoro: It looks like this piece of code is unnecessary! You already have your post up above.

index.js
```diff
@@ -47,8 +65,15 @@ server.post('/api/login', (req, res) => {
...
+            res.status(500).send(`failed to log you out`)
+        } else {
+            res.status(201).send('successfully logged you out')
+        }
+    })
 
-server.post('/api/login', (req, res) => {
 })
 
 server.listen(2525, () => console.log(`I'm alive!`))
```
johnoro: Typically we want a port variable, i.e.:
```javascript
const PORT = 2525;
server.listen(PORT, () => console.log(`I'm alive on port ${port}!`));
```

database/dbHelpers.js
```diff
@@ -0,0 +1,21 @@
...
+get: (users) => {
+    return db('users').select('id', 'username')
+},
+
+insert: (user) => {
+    return db('users').insert(user);
+},
+
+findByUsername: (username) => {
+    return db('users').where('username', username);
```
johnoro: You could use an object here to make the argument for your `where` a tad shorter:
```javascript
return db('users').where({ username });
```

PR #549, opened 2019-01-30
Mitchell Robles - auth-i

mitchellr92:
@johnoro


data/dbHelpers.js
```diff
@@ -0,0 +1,18 @@
...
+    getUsers: () => {
+        return db('users')
+    },
+
+    insertUser: (user) => {
+        return db('users').insert(user);
+    },
+
+    findByUsername: (username) => {
+        return db('users').where('username', username);
```
johnoro: You could use an object for the argument of `where` to make this a bit shorter:
```javascript
return db('users').where({ username });
```

index.js
```diff
@@ -0,0 +1,51 @@
...
+            } else {
+                res.status(404).json({ err: 'You shall not pass!' })
+            }
+        })
+        .catch(err => {
+            res.status(500).send(err);
+        })
+})
+
+server.listen(4400, () => console.log(`Listening on port ${PORT}`));
```
johnoro: You should use the PORT variable as the first argument for your `listen` here:
```javascript
server.listen(PORT, () => console.log(`Listening on port ${PORT}`)); 
```

index.js
```diff
@@ -0,0 +1,51 @@
...
+
+const PORT = 4400;
+
+server.get('/api/users', (req, res) => {
+    db.getUsers()
+        .then(users => {
+            res.json(users)
+        })
+        .catch(err => {
+            res.send(err)
```
johnoro: You should send an explicit status here; I'm pretty sure this defaults to `200` (OK) when you're probably wanting a `500` (server error)

index.js
```diff
@@ -0,0 +1,51 @@
...
+        });
+});
+
+server.post('/api/register', (req, res) => {
+    const user = req.body;
+    user.password = bcrypt.hashSync(user.password, 14);
+    db.insertUser(user)
+        .then(ids => {
+            res.status(201).json({ id: ids[0] });
```
johnoro: A couple of other options for getting this first id (`ids[0]`):

- `first` method to grab the first element of an array
```javascript
db.insertUser(user)
  .first()
  .then(id => {
    res.status(201).json({ id });
    // ...
```
- array destructuring in the parameter list to get the first element
```javascript
db.insertUser(user)
  .then(([ id ]) => {
    res.status(201).json({ id });
    // ...
```

index.js
```diff
@@ -34,10 +53,13 @@ server.post('/api/register', (req, res) => {
 });
 
 server.post('/api/login', (req, res) => {
+
+
     const bodyUser = req.body;
     db.findByUsername(bodyUser.username)
         .then(users => {
             if (users.length && bcrypt.compareSync(bodyUser.password, users[0].password)) {
+                req.session.userId = users[0].id;
```
johnoro: You could use the `first` method to make this slightly less clunky/more clear:
```javascript
const { body } = req;
db.findByUsername(body.username).first()
  .then(user => {
    if (user && bcrypt.compareSync(body.password, user.password)) {
      req.session.userId = user.id;
      // ...
```

PR #552, opened 2019-02-01
Courtney Buratto / auth-i

cocoitali:
@eddygonzalez9708

### Design and build the following endpoints:
- [x] POST /api/register:  Creates a `user` using the information sent inside the `body` of the request. **Hash the password** before saving the user to the database.                                                                                                                                                 
- [x] POST /api/login: Use the credentials sent inside the `body` to authenticate the user. On successful login, create a new session for the user and send back a 'Logged in' message and a cookie that contains the user id. If login fails, respond with the correct status code and the message: 'You shall not pass!' 
- [x] GET /api/users: If the user is logged in, respond with an array of all the users contained in the database. If the user is not logged in respond with the correct status code and the message: 'You shall not pass!'.            
- [x] After we cover the lecture on **sessions** and **cookies**, use them to keep a record of logged in users across requests.

## Stretch Problem


cocoitali:
^ above commit contains session info, saves user id for 1 day

PR #600, opened 2019-09-05
Kelsy wells

kelsywells:


data/auth-router.js
```diff
@@ -0,0 +1,41 @@
...
+      })
+      .catch(error => {
+        res.status(500).json(error);
+      });
+  });
+
+  router.post('/api/login', (req, res) => {
+    let { username, password } = req.body;
+  
+    password= bcrypt.hashSync(user.password, 10);
```
sibhat: You probably wanna use the plain password without hashing it, as bcrypt.compareSync() func will hash the password and compare to the second argument you passed in.

PR #680, opened 2020-03-16
updates instructions to one day project

luishrd:
This project used to be two days.

We are adding a Unit Testing module to this sprint and need to combine Modules 1 and 2.

The guided project will be updated accordingly.


zoelud:
Overall project looks good, I'd like to add more detail if we can, since we're already editing it. Any additional instructions we can give to improve student products would be awesome. See comments!

